PR #29, opened 2019-11-11
Add Firewall Rule

### LINKED ISSUE #24: GCP: Firewall Rule
This is the firewalls API: https://cloud.google.com/compute/docs/reference/rest/v1/firewalls

firewalls

deepanjan90:
Fixes #24: This PR adds the Firewall Resource.

### DIFF
--- a/src/main/java/gyro/google/compute/FirewallAllowDenyRule.java
+++ b/src/main/java/gyro/google/compute/FirewallAllowDenyRule.java
@@ -27,11 +27,12 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 public class FirewallAllowDenyRule extends Diffable {
-    private static final Pattern PORT_PATTERN = Pattern.compile("((?!(0))\\d+(-[1-9]\\d+)?)");
+    private static final Pattern PORT_PATTERN = Pattern.compile("(?<start>[1-9]\\d*)-?(?<end>[1-9]\\d*)?");
 
     private String protocol;
     private Set<String> ports;
@@ -115,12 +116,22 @@ public class FirewallAllowDenyRule extends Diffable {
         return errors;
     }
 
+    /**
+     * Checks if the given port string is valid or not.
+     * Valid ports are any positive numbers without leading zeroes, or a range specified with '-' with numbers on either side. The start number less that the end number and both numbers being positive without leading zeroes
+     * @param port The port string to check for validity
+     * @return true if valid false if invalid
+     */
     private boolean validatePort(String port) {
-        if (PORT_PATTERN.matcher(port).matches()) {
-            String[] split = port.split("-");
-            return split.length == 1 || (Integer.parseInt(split[0]) < Integer.parseInt(split[1]));
-        } else {
-            return false;
+        Matcher matcher = PORT_PATTERN.matcher(port);
+        if (matcher.matches()) {
+            Integer start = Integer.valueOf(matcher.group("start"));
+            Integer end = matcher.group("end") != null
+                ? Integer.parseInt(matcher.group("end")) : port.contains("-") ? -1 : 0;
+
+            return end == 0 || (start >= 0 && end >= 0 && start < end);
         }
+
+        return false;
     }
 }
--- a/src/main/java/gyro/google/compute/FirewallAllowDenyRule.java
+++ b/src/main/java/gyro/google/compute/FirewallAllowDenyRule.java
@@ -1,0 +1,110 @@
+package gyro.google.compute;
+
+import com.google.api.services.compute.model.Firewall.Allowed;
+import com.google.api.services.compute.model.Firewall.Denied;
+import gyro.core.resource.Diffable;
+import gyro.core.resource.Updatable;
+import gyro.core.validation.Required;
+import gyro.core.validation.ValidationError;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+public class FirewallAllowDenyRule extends Diffable {
+    private static final Pattern PORT_PATTERN = Pattern.compile("((?!(0))\\d+(-[1-9]\\d+)?)");
+
+    private String protocol;
+    private Set<String> ports;
+
+    /**
+     * The protocol that is going to be matched for the incoming/outgoing requests. Valid values are ``tcp``, ``udp``, ``icmp``, ``esp``, ``ah``, ``ipip``, ``sctp`` or a IP protocol number. (Required)
+     */
+    @Required
+    public String getProtocol() {
+        return protocol != null ? protocol.toLowerCase() : null;
+    }
+
+    public void setProtocol(String protocol) {
+        this.protocol = protocol;
+    }
+
+    /**
+     * A list of ports associated with the protocol. Can only be set when 'protocol' set to  ``tcp`` or ``udp``. Valid values are valid port number or port number range. Ex. ``22`` or ``22-443``.
+     */
+    @Updatable
+    public Set<String> getPorts() {
+        if (ports == null) {
+            ports = new HashSet<>();
+        }
+
+        return ports;
+    }
+
+    public void setPorts(Set<String> ports) {
+        this.ports = ports;
+    }
+
+    @Override
+    public String primaryKey() {
+        return getProtocol();
+    }
+
+    void copyFrom(Allowed allowed) {
+        setProtocol(allowed.getIPProtocol());
+        setPorts(allowed.getPorts() != null ? new HashSet<>(allowed.getPorts()) : null);
+    }
+
+    void copyFrom(Denied denied) {
+        setProtocol(denied.getIPProtocol());
+        setPorts(denied.getPorts() != null ? new HashSet<>(denied.getPorts()) : null);
+    }
+
+    Allowed toAllowed() {
+        Allowed allowed = new Allowed();
+        allowed.setIPProtocol(getProtocol());
+        if (!getPorts().isEmpty()) {
+            allowed.setPorts(new ArrayList<>(getPorts()));
+        }
+        return allowed;
+    }
+
+    Denied toDenied() {
+        Denied denied = new Denied();
+        denied.setIPProtocol(getProtocol());
+        if (!getPorts().isEmpty()) {
+            denied.setPorts(new ArrayList<>(getPorts()));
+        }
+        return denied;
+    }
+
+    @Override
+    public List<ValidationError> validate() {
+        List<ValidationError> errors = new ArrayList<>();
+
+        if (!getProtocol().equals("tcp") && !getProtocol().equals("udp") && !getPorts().isEmpty()) {
+            errors.add(new ValidationError(this, "ports", "'ports' can only be set when 'protocol' is set to either 'tcp' or 'udp'"));
+        }
+
+        List<String> invalidPorts = getPorts().stream().filter(port -> !validatePort(port)).collect(Collectors.toList());
+        if (!invalidPorts.isEmpty()) {
+            for (String port : invalidPorts) {
+                errors.add(new ValidationError(this, "ports", String.format("invalid entry %s. Must be an integer or a valid range", port)));
+            }
+        }
+
+        return errors;
+    }
+
+    private boolean validatePort(String port) {
+        if (PORT_PATTERN.matcher(port).matches()) {
+            String[] split = port.split("-");
+            return split.length == 1 || (Integer.parseInt(split[0]) < Integer.parseInt(split[1]));
+        } else {
+            return false;
+        }
+    }
+}
--- a/src/main/java/gyro/google/compute/FirewallAllowDenyRule.java
+++ b/src/main/java/gyro/google/compute/FirewallAllowDenyRule.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2019, Perfect Sense, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package gyro.google.compute;
 
 import com.google.api.services.compute.model.Firewall.Allowed;
--- a/src/main/java/gyro/google/compute/FirewallResource.java
+++ b/src/main/java/gyro/google/compute/FirewallResource.java
@@ -61,7 +61,7 @@ import java.util.stream.Collectors;
  *
  *         rule-type: "allow"
  *
- *         rule
+ *         allowed
  *             protocol: "tcp"
  *             ports: ["95-96", "80-85"]
  *         end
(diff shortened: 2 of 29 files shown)


src/main/java/gyro/google/compute/FirewallAllowDenyRule.java
```diff
@@ -0,0 +1,126 @@
...
+        return errors;
+    }
+
+    private boolean validatePort(String port) {
+        if (PORT_PATTERN.matcher(port).matches()) {
+            String[] split = port.split("-");
+            return split.length == 1 || (Integer.parseInt(split[0]) < Integer.parseInt(split[1]));
+        } else {
+            return false;
+        }
```
beetlebugorg: If you update to the regex above with named groups then this logic can be simplified:

```java
Matcher matcher = PORT_PATTERN.matcher(port);
if (matcher.matches()) {
    Integer start = Integer.valueOf(matcher.group("start"));
    Integer end = matcher.group("end") != null ? Integer.valueOf(matcher.group("end")) : 0;

    return end == 0 || start < end;
}

return false;
```

src/main/java/gyro/google/compute/FirewallRuleResource.java
```diff
@@ -0,0 +1,580 @@
...
+ *         end
+ *
+ *         source-tags: [
+ *             "source-tag-example"
+ *         ]
+ *     end
+ */
+@Type("firewall-rule")
+public class FirewallRuleResource extends ComputeResource implements Copyable<Firewall> {
+    private static final Pattern NAME_PATTERN = Pattern.compile("[a-z]([-a-z0-9]*[a-z0-9])?");
```
beetlebugorg: I don't think this will perfectly match valid names. It should check for length. This variant ignores a trailing `-`. I think we should detect that and error. 

How does the Google UI and API handle uppercase characters? Does it automatically lowercase them?

From the docs:

> The name must be 1-63 characters long and match the regular expression `a-z?. The first character must be a lowercase letter, and all following characters (except for the last character) must be a dash, lowercase letter, or digit. The last character must be a lowercase letter or digit.

src/main/java/gyro/google/compute/FirewallRuleResource.java
```diff
@@ -0,0 +1,580 @@
...
+    private Set<String> sourceTags;
+    private Boolean logConfig;
+    private Set<String> targetServiceAccounts;
+    private Set<String> targetTags;
+
+    // Read-only
+    private String id;
+    private String selfLink;
+
+    private enum ACTION_TYPE {PATCH, UPDATE, CREATE}
```
beetlebugorg:
```suggestion
    private enum ACTION_TYPE {
        PATCH, 
        UPDATE, 
        CREATE
    }
```

src/main/java/gyro/google/compute/FirewallRuleResource.java
```diff
@@ -0,0 +1,580 @@
...
+
+    /**
+     * The name of the firewall rule. Needs to follow Google firewall rule naming convention. (Required)
```
beetlebugorg: Should explain the naming rules.

src/main/java/gyro/google/compute/FirewallRuleResource.java
```diff
@@ -0,0 +1,580 @@
...
+     * Fields 'destination ranges', 'disabled', 'priority', 'source service accounts', 'log config', 'target service account' and 'denied' can only be updated using the patch api.
+     *
+     * @param client The compute client
+     * @param actionType The type of action (CREATE, UPDATE or PATCH)
+     * @return Firewall object
+     */
+    private Firewall toFirewall(Compute client, ACTION_TYPE actionType) {
+        Firewall firewall = getFirewall(client, actionType);
+
+        if (actionType.equals(ACTION_TYPE.UPDATE) || actionType.equals(ACTION_TYPE.CREATE)) {
```
beetlebugorg:
```suggestion
        if (actionType == ACTION_TYPE.UPDATE || actionType == ACTION_TYPE.CREATE) {
```

src/main/java/gyro/google/compute/FirewallRuleResource.java
```diff
@@ -0,0 +1,580 @@
...
+            firewall.setSourceRanges(!getSourceRanges().isEmpty() ? new ArrayList<>(getSourceRanges()) : null);
+            firewall.setSourceTags(!getSourceTags().isEmpty() ? new ArrayList<>(getSourceTags()) : null);
+            firewall.setTargetTags(!getTargetTags().isEmpty() ? new ArrayList<>(getTargetTags()) : null);
+
+            if (getRuleType().equals("ALLOW")) {
+                firewall.setAllowed(getRule().stream().map(FirewallAllowDenyRule::toAllowed).collect(Collectors.toList()));
+            }
+        }
+
+        if (actionType.equals(ACTION_TYPE.PATCH) || actionType.equals(ACTION_TYPE.CREATE)) {
```
beetlebugorg:
```suggestion
        if (actionType == ACTION_TYPE.PATCH || actionType == ACTION_TYPE.CREATE) {
```

src/main/java/gyro/google/compute/FirewallRuleResource.java
```diff
@@ -0,0 +1,580 @@
...
+
+        try {
+            // update
+            operation = client.firewalls().update(getProjectId(), getName(), toFirewall(client, ACTION_TYPE.UPDATE)).execute();
+            error = waitForCompletion(client, operation);
+            if (error != null) {
+                throw new GyroException(error.toPrettyString());
+            }
+
+            // patch
```
beetlebugorg: I'm not sure I follow the need for using both update and patch. Isn't patch just an optimized variant of update?

https://cloud.google.com/compute/docs/api/how-tos/performance#partial

src/main/java/gyro/google/compute/FirewallRuleResource.java
```diff
@@ -0,0 +1,580 @@
...
+    private Firewall toFirewall(Compute client, ACTION_TYPE actionType) {
+        Firewall firewall = getFirewall(client, actionType);
+
+        if (actionType.equals(ACTION_TYPE.UPDATE) || actionType.equals(ACTION_TYPE.CREATE)) {
+            firewall.setDescription(getDescription());
+            firewall.setSourceRanges(!getSourceRanges().isEmpty() ? new ArrayList<>(getSourceRanges()) : null);
+            firewall.setSourceTags(!getSourceTags().isEmpty() ? new ArrayList<>(getSourceTags()) : null);
+            firewall.setTargetTags(!getTargetTags().isEmpty() ? new ArrayList<>(getTargetTags()) : null);
+
+            if (getRuleType().equals("ALLOW")) {
```
beetlebugorg: I noticed that for `ACTION_TYPE.UPDATE` you process `ALLOW` and in the following block for `ACTION_TYPE.PATCH` you process `DENY`. 

It seems like something is wrong with the use of the APIs.

src/main/java/gyro/google/compute/FirewallRuleResource.java
deepanjan90: Google UI and API both do not allow upper case and errors

Updated by removing the validation and letting the Google API throw the exception that also specify the constraints.

src/main/java/gyro/google/compute/FirewallRuleResource.java
deepanjan90: The update api allows for certain fields to be updated, for the rest the patch needs to be used. As per their docs they specify the ones `update` supports:
```
/**
     * Updates the specified firewall rule with the data included in the request. The PUT method can
     * only update the following fields of firewall rule: allowed, description, sourceRanges,
     * sourceTags, targetTags.
     *
     *
```

src/main/java/gyro/google/compute/FirewallRuleResource.java
deepanjan90: As discussed above, the update API allows updating `allowed` while the patch API allows updating `denied`.

src/main/java/gyro/google/compute/FirewallRuleResource.java
deepanjan90: We figured that patch is able to update all the fields and so the `gyro update` now uses patch to update all the fields for the firewall rule.

src/main/java/gyro/google/compute/FirewallRuleResource.java
deepanjan90: We figured that `patch` is able to update all the fields and so the `gyro update` now uses patch to update all the fields for the firewall rule. So the above code is replaced

src/main/java/gyro/google/compute/FirewallResource.java
```diff
@@ -0,0 +1,546 @@
...
+ *         ]
+ *     end
+ */
+@Type("firewall-rule")
+public class FirewallResource extends ComputeResource implements Copyable<Firewall> {
+    private String name;
+    private NetworkResource network;
+    private String description;
+    private String ruleType;
+    private Set<FirewallAllowDenyRule> rule;
```
beetlebugorg: This doesn't follow the GCP data model. In their model they have an `allowed` list and `denied` list. I realize the model of those two lists is the same but we should mirror the underlying API so we're consistent.

src/main/java/gyro/google/compute/FirewallResource.java
```diff
@@ -0,0 +1,546 @@
...
+    private Set<String> destinationRanges;
```
beetlebugorg: We should determine if these needs to be `Set`s or not. I'm worried that blindly making them sets will cause problems in the future when order matters.

Unless the API returns the values in a different order than we provided them, these should remain `List`s as the underlying model is a `List`.

src/main/java/gyro/google/compute/FirewallResource.java
```diff
@@ -0,0 +1,546 @@
...
+        return firewall;
+    }
+
+    private Firewall getFirewall(Compute client) {
+        Firewall firewall = null;
+
+        try {
+            firewall = client.firewalls().get(getProjectId(), getName()).execute();
+        } catch (GoogleJsonResponseException je) {
+            if (!je.getDetails().getMessage().matches("The resource (.*) was not found")) {
```
beetlebugorg: Does the error details no include an error code you can check?

src/main/java/gyro/google/compute/FirewallResource.java
deepanjan90: Except for the allowed and denied attributes, all other collection type attributes are treated as sets, as the API does not complain for duplicates, but internally removes them, so using sets handles the case on gyro end as well, where duplicates are automatically removed.

src/main/java/gyro/google/compute/FirewallResource.java
```diff
@@ -0,0 +1,572 @@
...
+
+        if (getRuleType().equals("ALLOW") && getAllowed().isEmpty()) {
+            errors.add(new ValidationError(this, "allowed", "'allowed' needs to be set when 'rule-type' set to 'ALLOW'."));
+        }
+
+        if (getRuleType().equals("DENY") && getDenied().isEmpty()) {
+            errors.add(new ValidationError(this, "denied", "'denied' needs to be set when 'rule-type' set to 'DENY'."));
+        }
+
+        if (getDirection().equals("INGRESS")) {
```
beetlebugorg: The hope was to use the new `@ConflictsWith` annotation to get rid of this entire section of validation. See my suggestion above and let me know if that doesn't work.

src/main/java/gyro/google/compute/FirewallResource.java
```diff
@@ -0,0 +1,572 @@
...
+
+    public void setDenied(List<FirewallDenied> denied) {
+        this.denied = denied;
+    }
+
+    /**
+     * A set of destination IP in cidr form that the firewall rule applies to. Can only be set when 'direction' set to 'EGRESS'.
+     */
+    @Updatable
+    public Set<String> getDestinationRanges() {
```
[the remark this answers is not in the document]
deepanjan90: This needs a value based validation https://github.com/perfectsense/gyro-google-provider/blob/feature/firewall-rules/src/main/java/gyro/google/compute/FirewallResource.java#L499-L503

In combination with https://github.com/perfectsense/gyro-google-provider/blob/feature/firewall-rules/src/main/java/gyro/google/compute/FirewallResource.java#L509-L523, it automatically avoids the conflicting scenario.

src/main/java/gyro/google/compute/FirewallResource.java
```diff
@@ -0,0 +1,572 @@
...
+
+    public void setPriority(Integer priority) {
+        this.priority = priority;
+    }
+
+    /**
+     * A set of source IP in cidr form that the firewall rule applies to. Can only be set when 'direction' set to 'INGRESS'.
+     */
+    @Updatable
+    public Set<String> getSourceRanges() {
```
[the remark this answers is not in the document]
deepanjan90: This needs a value based validation https://github.com/perfectsense/gyro-google-provider/blob/feature/firewall-rules/src/main/java/gyro/google/compute/FirewallResource.java#L509-L511

In combination with https://github.com/perfectsense/gyro-google-provider/blob/feature/firewall-rules/src/main/java/gyro/google/compute/FirewallResource.java#L500-L506, it automatically avoids the conflicting scenario.

src/main/java/gyro/google/compute/FirewallResource.java
```diff
@@ -0,0 +1,572 @@
...
+
+    public void setSourceRanges(Set<String> sourceRanges) {
+        this.sourceRanges = sourceRanges;
+    }
+
+    /**
+     * A set of service accounts that the incoming requests are going to be matched with only if it originated from instances of the accounts specified.  Can only be set when 'direction' set to 'INGRESS'. Only one of 'source-service-account' or 'source-tags' can be set.
+     */
+    @Updatable
+    @ConflictsWith({"source-tags", "target-tags"})
```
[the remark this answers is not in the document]
deepanjan90: This value based validation handles the 'destination-ranges' attribute https://github.com/perfectsense/gyro-google-provider/blob/feature/firewall-rules/src/main/java/gyro/google/compute/FirewallResource.java#L500-L502. This avoids the conflicting case. Should it still be part of the annotation ?

src/main/java/gyro/google/compute/FirewallResource.java
```diff
@@ -0,0 +1,572 @@
...
+
+    public void setSourceServiceAccounts(Set<String> sourceServiceAccounts) {
+        this.sourceServiceAccounts = sourceServiceAccounts;
+    }
+
+    /**
+     * A set of tags that the incoming requests are going to be matched with only if it originated from instances whose primary network interface has the same tags. Can only be set when 'direction' set to 'INGRESS'. Only one of 'source-service-account' or 'source-tags' can be set.
+     */
+    @Updatable
+    public Set<String> getSourceTags() {
```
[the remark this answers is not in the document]
deepanjan90: The value based validation https://github.com/perfectsense/gyro-google-provider/blob/feature/firewall-rules/src/main/java/gyro/google/compute/FirewallResource.java#L521-L523 in combination to https://github.com/perfectsense/gyro-google-provider/blob/feature/firewall-rules/src/main/java/gyro/google/compute/FirewallResource.java#L513-L515 handles this, should it still have the `@ConflictsWith` annotation ?

src/main/java/gyro/google/compute/FirewallResource.java
deepanjan90: The `@ConflictsWith` can handle the conflicts arising from these validated attributes, but a value based validation will still be needed. The value of the `direction` attribute either `INGRESS` or `EGRESS` dictates which fields are allowed, which handles the conflicts as the conflicting fields belong to different values for the attribute `direction`.

We could create an annotation like `@AllowedIf(field='direction', value={'INGRESS'})` which validates a field based on what value is configured on another field. This will get rid of this whole block of validating separately.

src/main/java/gyro/google/compute/FirewallResource.java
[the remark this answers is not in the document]
beetlebugorg: I think you're arguing that your method of validation provides a better error message because other than that I don't see how my suggestion doesn't work.

How about this suggested changed to the `@Required` annotation to allow for a more specific error message?

```java
@Updatable
@Required(when = "direction", equals = "INGRESS", constraint = RequiredConstraint.ONLY_ONE)
public Set<String> getSourceRanges() {
...
}
```

The idea is that you'd annotate all fields that are potentially required when `direction` is set to `INGRESS`. The constraint determines if one field should be set, if all fields should be set, or if any field cannot be set.

Thoughts?

PR #32, opened 2019-11-13
Refactor google provider package name

deepanjan90:
Rename provider package name from `gyro.google` to `gyro.provider.google`

Add package info to `compute` package


deepanjan90:
Package name refactor may be part of the refactor using the autogen code.

Issue #22, opened 2019-11-01
GCP: VM instance


kluman:
> Compute Engine lets you use virtual machines that run on Google's infrastructure. Create micro-VMs or larger instances running Debian, Windows, or other standard images. Create your first VM instance, import it using a migration service, or try the quickstart to build a sample app.

This is the instances API: https://cloud.google.com/compute/docs/reference/rest/v1/instances

<img width="1132" alt="instances" src="https://user-images.githubusercontent.com/286670/68052739-b4474200-fcc0-11e9-81d5-3a72c792f7a5.png">

deepanjan90:
Add reference to the following resource:
 - Route Resource (`nextHopInstance`)
 - Network Endpoint Resource (`endpointGroup` --> `instance`)

PR #37, opened 2019-11-14
Add Network Endpoint Group

### LINKED ISSUE #15: GCP: Network Endpoint Group
> A network endpoint group (NEG) is a collection of IP addresses that you can apply networking functions to, like load balancing, firewalls and logging. If you're creating an HTTP(S) load balancer with backend containers, your NEG will be created automatically.

This uses the network endpoints groups API (beta): https://cloud.google.com/compute/docs/reference/rest/beta/networkEndpointGroups

network-endpoint-groups

deepanjan90:
Fixes #15: This PR adds the Network Endpoint Resource. Also adds a skeleton for Instance Resource so that the Network Endpoint Resource can reference it.

Instance Resource needs to be implemented as part of #22

### DIFF
--- a/src/main/java/gyro/google/compute/NetworkEndpointGroupResource.java
+++ b/src/main/java/gyro/google/compute/NetworkEndpointGroupResource.java
@@ -40,6 +40,10 @@ import java.util.stream.Collectors;
  *         subnet: $(google::subnet subnet-example-network-endpoint-group)
  *         default-port: 80
  *         zone: "us-east1-b"
+ *
+ *         endpoint
+ *             instance: "instance-example"
+ *         end
  *     end
  */
 @Type("network-endpoint-group")
--- a/src/main/java/gyro/google/compute/NetworkEndpointGroupResource.java
+++ b/src/main/java/gyro/google/compute/NetworkEndpointGroupResource.java
@@ -1,0 +1,353 @@
+package gyro.google.compute;
+
+import com.google.api.client.googleapis.json.GoogleJsonResponseException;
+import com.google.api.services.compute.Compute;
+import com.google.api.services.compute.model.NetworkEndpointGroup;
+import com.google.api.services.compute.model.NetworkEndpointGroupsAttachEndpointsRequest;
+import com.google.api.services.compute.model.NetworkEndpointGroupsDetachEndpointsRequest;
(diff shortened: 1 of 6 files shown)


src/main/java/gyro/google/compute/NetworkEndpointGroupResource.java
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    /**
+     * The fully qualified url of the network endpoint group.
+     */
+    @Output
```
deepanjan90:
```suggestion
    @Id
    @Output
```

src/main/java/gyro/google/compute/NetworkEndpointResource.java
```diff
@@ -0,0 +1,119 @@
...
+        if (healthStatus == null) {
+            healthStatus = new ArrayList<>();
+        }
+
+        return healthStatus;
+    }
+
+    public void setHealthStatus(List<String> healthStatus) {
+        this.healthStatus = healthStatus;
+    }
```
deepanjan90: Docs

src/main/java/gyro/google/compute/NetworkEndpointResource.java
```diff
@@ -0,0 +1,119 @@
...
+public class NetworkEndpointResource extends Diffable implements Copyable<NetworkEndpointWithHealthStatus> {
+
+    private InstanceResource instance;
+    private Integer port;
+    private String ipAddress;
+
+    // Read-only
+    private List<String> healthStatus;
+
+    @Id
```
deepanjan90: No @Id annotation for a subresource, as findById only works for resources
```suggestion
```

src/main/java/gyro/google/compute/NetworkEndpointGroupResource.java
kluman: Done.

src/main/java/gyro/google/compute/NetworkEndpointResource.java
kluman: Done.

src/main/java/gyro/google/compute/NetworkEndpointResource.java
kluman: Done.

PR #44, opened 2019-11-21
Add Routes

### LINKED ISSUE #23: GCP: Routes
This is the routes API: https://cloud.google.com/compute/docs/reference/rest/v1/routes

routes

deepanjan90:
Fixes #23: This PR adds the Firewall Resource.


src/main/java/gyro/google/compute/RouteResource.java
```diff
@@ -0,0 +1,345 @@
...
+    private Set<String> tags;
+    private String nextHopGateway;
+    private String nextHopVpnTunnel;
+    private InstanceResource nextHopInstance;
+    private String nextHopIp;
+
+    // Read-only
+    private String id;
+    private String selfLink;
+    private List<RouteWarning> warnings;
```
beetlebugorg: I think these warnings should be output after a create or update rather than stored.

PR #51, opened 2019-11-27
DNS zones, record sets, and policies.

### LINKED ISSUE #14: GCP: DNS Zones and policies
> DNS zones let you define your namespace. You can create public or private zones.

This uses the Cloud DNS API: https://cloud.google.com/dns/docs/reference/v1/

dns-zones

sshyub:
Fix #14


src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+    private Boolean enableLogging;
+
+    private String id;
+
+    private String name;
+
+    private List<DnsPolicyNetwork> network;
+
+    /**
+     * Sets an alternative name server for the associated networks. When specified, all DNS queries are forwarded to a name server that you choose. Names such as .internal are not available when an alternative name server is specified. The value may be ``null``.
```
deepanjan90: This doc is to help a user configure the value for the field. As someone cannot set ``null`` in the config this portion is not necessary. Unless a field is required, it is optional and so if not configured by a user, will behave accordingly.

Also by convention we try to reword the docs a bit, and not directly use what  is there in the API.
```suggestion
     * Sets an alternative name server for the associated networks. When specified, all DNS queries are forwarded to a name server that you choose. Names such as .internal are not available when an alternative name server is specified.
```

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    public void setAlternativeNameServerConfig(DnsPolicyAlternativeNameServerConfig alternativeNameServerConfig) {
+        this.alternativeNameServerConfig = alternativeNameServerConfig;
+    }
+
+    /**
+     * A mutable string of at most 1024 characters associated with this resource for the user's convenience. Has no effect on the policy's function. The value may be ``null``.
+     */
+    @Required
```
deepanjan90: If the attribute is required then it cannot be null.

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+    public String getDescription() {
+        return description;
+    }
+
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
+    /**
+     * Allows networks bound to this policy to receive DNS queries sent by VMs or applications over VPN connections. When enabled, a virtual IP address will be allocated from each of the sub- networks that are bound to this policy. The value may be ``null``.
```
deepanjan90: This doc is to help a user configure the value for the field. As someone cannot set null in the config this portion is not necessary. Unless a field is required, it is optional and so if not configured by a user, will behave accordingly.
```suggestion
     * Allows networks bound to this policy to receive DNS queries sent by VMs or applications over VPN connections. When enabled, a virtual IP address will be allocated from each of the sub- networks that are bound to this policy.
```

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+    public Boolean getEnableInboundForwarding() {
+        return enableInboundForwarding;
+    }
+
+    public void setEnableInboundForwarding(Boolean enableInboundForwarding) {
+        this.enableInboundForwarding = enableInboundForwarding;
+    }
+
+    /**
+     * Controls whether logging is enabled for the networks bound to this policy. Defaults to no logging if not set. The value may be ``null``.
```
deepanjan90: This doc is to help a user configure the value for the field. As someone cannot set null in the config this portion is not necessary. Unless a field is required, it is optional and so if not configured by a user, will behave accordingly.

Also by convention we try to reword the docs a bit and not use the one directly from the API. The docs are generally reworded in a way so that the language and sentence construction is consistent throughout our documentation. 
```suggestion
     * When enabled, turns on logging for the networks bound to this policy. Defaults to ``false``.
```

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+    public Boolean getEnableLogging() {
+        return enableLogging;
+    }
+
+    public void setEnableLogging(Boolean enableLogging) {
+        this.enableLogging = enableLogging;
+    }
+
+    /**
+     * Unique identifier for the resource; defined by the server (output only). The value may be ``null``.
```
deepanjan90: This doc is to help a user configure the value for the field. As someone cannot set null in the config this portion is not necessary. Unless a field is required, it is optional and so if not configured by a user, will behave accordingly.

For an output variable the annotation `@Output` makes the variable be written in a separate section in the documentation and mentioning in the doc that it is an `Output variable` is not necessary.

Also by convention we try to reword the docs a bit and not use the one directly from the API. The docs are generally reworded in a way so that the language and sentence construction is consistent throughout our documentation.

```suggestion
     * The generated ID for the policy.
```

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    /**
+     * User assigned name for this policy. The value may be ``null``.
+     */
+    @Required
```
deepanjan90: If the value is required then it cannot be null

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    /**
+     * List of network names specifying networks to which this policy is applied. The value may be ``null``.
```
deepanjan90:
```suggestion
     * List of network names specifying networks to which this policy is applied.
```

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+
+    public void setNetwork(List<DnsPolicyNetwork> network) {
+        this.network = network;
+    }
+
+    @Override
+    public boolean doRefresh() throws Exception {
+        Dns client = createClient(Dns.class);
+        Policy policy = client.policies().get(getProjectId(), getName()).execute();
+        return refreshFrom(policy);
```
deepanjan90: Following convention.
```suggestion
        
        return refreshFrom(policy);
```

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+                    return diffablePolicyNetwork;
+                })
+                .collect(Collectors.toList()));
+        }
+    }
+
+    private boolean refreshFrom(Policy managedZone) {
+        copyFrom(managedZone);
+        return true;
+    }
```
deepanjan90: This method does not serve any special purpose and should be removed and the `copyFrom` method should be used in all the places this method is used.

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+            } else if (changedFieldName.equals("network")) {
+                List<DnsPolicyNetwork> networks = getNetwork();
+                policy.setNetworks(networks == null
+                    ? Data.nullOf(ArrayList.class)
+                    : networks.stream()
+                        .map(DnsPolicyNetwork::copyTo)
+                        .collect(Collectors.toList()));
+            }
+        }
+        Dns client = createClient(Dns.class);
```
deepanjan90: Conventionally we do declare the client at the top.

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+        Policy response = client.policies().create(getProjectId(), policy).execute();
+        refreshFrom(response);
+    }
+
+    @Override
+    public void doUpdate(
+        GyroUI ui, State state, Resource current, Set<String> changedFieldNames) throws Exception {
+        if (!(current instanceof PolicyResource)) {
+            throw new GyroException("Incompatible resource type! " + current.getClass().getName());
+        }
```
deepanjan90: Gyro would not be calling the `update()` with the wrong  `current` resource and the check is redundant

src/main/java/gyro/google/dns/PolicyResource.java
```diff
@@ -0,0 +1,276 @@
...
+
+    /**
+     * List of network names specifying networks to which this policy is applied. The value may be ``null``.
+     *
+     * @subresource gyro.google.dns.DnsPolicyNetwork
+     */
+    // XXX: https://github.com/perfectsense/gyro/issues/190
+    @Updatable
+    public List<DnsPolicyNetwork> getNetwork() {
+        return network;
```
deepanjan90: Following convention of how we handle lists. This also helps eliminate null checks when using this getter.
```suggestion
        if (network == null) {
            network = new ArrayList<>();
        }
        
        return network;
```

src/main/java/gyro/google/dns/DnsPolicyTargetNameServer.java
```diff
@@ -0,0 +1,71 @@
...
+public class DnsPolicyTargetNameServer extends Diffable
+    implements Copyable<PolicyAlternativeNameServerConfigTargetNameServer>,
+    Requestable<PolicyAlternativeNameServerConfigTargetNameServer> {
+
+    private String ipv4Address;
+
+    /**
+     * IPv4 address to forward to. The value may be ``null``.
+     */
+    @Required
```
deepanjan90: Cannot be null if it is required.

src/main/java/gyro/google/dns/DnsPolicyTargetNameServer.java
```diff
@@ -0,0 +1,71 @@
...
+    @Updatable
```
deepanjan90: Should not be updatable, as the primary key uses this fields value.

src/main/java/gyro/google/dns/DnsPolicyAlternativeNameServerConfig.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private List<DnsPolicyTargetNameServer> targetNameServer;
+
+    /**
+     * Sets an alternative name server for the associated networks. When specified, all DNS queries are forwarded to a name server that you choose. Names such as .internal are not available when an alternative name server is specified. The value may be ``null``.
+     *
+     * @subresource gyro.google.dns.DnsPolicyTargetNameServer
+     */
+    public List<DnsPolicyTargetNameServer> getTargetNameServer() {
+        return targetNameServer;
```
deepanjan90: Following convention of how we handle lists. This also helps eliminate null checks when using this getter.

```suggestion
        if (targetNameServer == null) {
            targetNameServer = new ArrayList<>();
        }
        
        return targetNameServer;
```

src/main/java/gyro/google/dns/DnsPolicyNetwork.java
```diff
@@ -0,0 +1,80 @@
...
+    Requestable<PolicyNetwork> {
+
+    private NetworkResource network;
+
+    /**
+     * The VPC network to bind to. The value may be ``null``.
+     *
+     * @resource gyro.google.compute.NetworkResource
+     */
+    public NetworkResource getNetwork() {
```
deepanjan90: Can you create a DnsPolicyNetwork with no network assigned ?

src/main/java/gyro/google/dns/DnsPolicyNetwork.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+
+    @Override
+    public PolicyNetwork copyTo() {
+        PolicyNetwork policyNetwork = new PolicyNetwork();
+        NetworkResource network = getNetwork();
+
+        if (network != null) {
+            policyNetwork.setNetworkUrl(network.getSelfLink());
```
deepanjan90: The NetworkResource has the `@Id` field set on the `name` attribute. So while configuring this complex type if network is assigned a string, which means a network which is not managed by gyro, it will create an empty NetworkResource model and set the `name` attribute to value being passed. Correspondingly the `self-link` field will be null and cause this to fail.

The `@Id` field of the NetworkResource should instead be put on the `self-link` attribute, so if an unmanaged network is assigned to the `network` attribute of this complex type, and empty `NetworkResource` object is created with the `self-link` attribute set to be used.

This change will mean that in `copyFrom` the networkUrl is directly used by the `findById` as `self-link`aka the networkUrl is the `@Id` to identify the `NetworkResource` if present.

This change will also mean that all the places where `NetworkResource` is used like https://github.com/perfectsense/gyro-google-provider/blob/feature/dns/src/main/java/gyro/google/compute/SubnetworkResource.java#L199 will now be using the `self-link` field as is used in this complex-type as well their respective `copyFrom()` method needs to set the networkUrl directly and not strip out the name and use that like https://github.com/perfectsense/gyro-google-provider/blob/feature/dns/src/main/java/gyro/google/compute/SubnetworkResource.java#L177-L179

src/main/java/gyro/google/dns/PolicyResource.java
sshyub: I think declaring variables closer to where they are used is better as it's easier to understand and organize, and there's a chance that variables never get used if they are declared too early.
Also if you run through a some kind of linter like sonar, it will complain that it is declared too early.
so I rather change the convention to move variables close to where they are being used.

src/main/java/gyro/google/dns/PolicyResource.java
sshyub: Could you point me where types are checked?

src/main/java/gyro/google/dns/PolicyResource.java
sshyub: I see your point, but honestly I wouldn't be bothered if those kinda convention can't be automated by checkstyle or some tools since there's no way we can enforce them 100%, which could lead to wasting of time and slow down the reviewing process.
If you still think those conventions should be enforced, please capture it in README or in some place so that everyone can follow the rules.

examples/dns/managed-zone.gyro
```diff
@@ -0,0 +1,21 @@
+google::dns-managed-zone managed-zone-example-private
+    name: "managed-zone-example-private"
+    description: "Private Managed Zone Example"
+    dns-name: "private.example.com."
+    visibility: "private"
+
+    private-visibility-config
+        network
+            network: $(google::compute-network network-example)
```
jpencola: This network subresource should be defined within this gyro config. It should also be named contextually to this example. Something like:
```

google::compute-network network-example
    name: "managed-zone-vpc-example"
    description: "managed-zone-vpc-example-desc"
    routing-mode: "Regional"
end
```

src/main/java/gyro/google/dns/ManagedZoneResource.java
```diff
@@ -0,0 +1,405 @@
...
+ *         dnssec-config
+ *             state: "on"
+ *         end
+ *     end
+ */
+@Type("dns-managed-zone")
+public class ManagedZoneResource extends GoogleResource implements Copyable<ManagedZone> {
+
+    private String description;
+
```
jpencola: I don't think the blank lines between each field are consistent with our code style.

src/main/java/gyro/google/dns/ManagedZoneResource.java
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    public void setPrivateVisibilityConfig(ZonePrivateVisibilityConfig privateVisibilityConfig) {
+        this.privateVisibilityConfig = privateVisibilityConfig;
+    }
+
+    /**
+     * The zone's visibility: public zones are exposed to the Internet, while private zones are visible only to Virtual Private Cloud resources.
+     */
+    @ValidStrings({ "private", "public" })
```
jpencola: Shouldn't the visibility be updatable? For example if I create a managed zone as private with a private visibility config. But then update it by removing that config and setting to "public" this operation is skipped when using `gyro up`.

examples/dns/policy.gyro
```diff
@@ -0,0 +1,7 @@
+google::dns-policy dns-policy-example
+    name: "dns-policy-example"
+    description: "DNS Policy Example"
+    network
+        network: $(google::compute-network network-example)
```
jpencola: This network subresource should be defined within this gyro config. It should also be named contextually to this example. Something like:

```
google::compute-network network-example
    name: "policy-vpc-example"
    description: "policy-vpc-example-desc"
    routing-mode: "Regional"
end
```

src/main/java/gyro/google/dns/ResourceRecordSetResource.java
```diff
@@ -0,0 +1,240 @@
...
+ * Creates a Resource Record Set.
+ *
+ * Example
+ * -------
+ *
+ * .. code-block:: gyro
+ *
+ *     google::dns-resource-record-set managed-zone-example-public-www-cname
+ *         managed-zone: $(google::dns-managed-zone managed-zone-example-public)
+ *
```
jpencola: remove blank line (also remove from gyro config).

src/main/java/gyro/google/dns/ResourceRecordSetResource.java
```diff
@@ -0,0 +1,240 @@
...
+ *         name: "www.p.example.com."
+ *         type: "CNAME"
+ *         rrdatas: [ "fe.p.example.com." ]
+ *     end
+ *
+ *     google::dns-resource-record-set managed-zone-example-public-fe-a
+ *         managed-zone: $(google::dns-managed-zone managed-zone-example-public)
+ *
```
jpencola: remove blank line (also remove from gyro config).

src/main/java/gyro/google/dns/ResourceRecordSetResource.java
```diff
@@ -0,0 +1,240 @@
...
+    public ManagedZoneResource getManagedZone() {
+        return managedZone;
+    }
+
+    public void setManagedZone(ManagedZoneResource managedZone) {
+        this.managedZone = managedZone;
+    }
+
+    /**
+     * For example, www.example.com.
```
jpencola: Prefix this with `The name of the resource record set.`

src/main/java/gyro/google/dns/ResourceRecordSetResource.java
```diff
@@ -0,0 +1,240 @@
...
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    /**
+     * As defined in RFC 1035 (section 5) and RFC 1034 (section 3.6.1) -- see examples.
```
jpencola: Is there another way to convey what this field is in a concise way instead of relying only on the RFC doc?

src/main/java/gyro/google/dns/ResourceRecordSetResource.java
```diff
@@ -0,0 +1,240 @@
...
+        }
+        return rrdatas;
+    }
+
+    public void setRrdatas(List<String> rrdatas) {
+        this.rrdatas = rrdatas;
+    }
+
+    /**
+     * As defined in RFC 4034 (section 3.2).
```
jpencola: Is there another way to convey what this field is in a concise way instead of relying only on the RFC doc?

src/main/java/gyro/google/dns/ResourceRecordSetResource.java
```diff
@@ -0,0 +1,240 @@
...
+        }
+        return signatureRrdatas;
+    }
+
+    public void setSignatureRrdatas(List<String> signatureRrdatas) {
+        this.signatureRrdatas = signatureRrdatas;
+    }
+
+    /**
+     * Number of seconds that this ResourceRecordSet can be cached by resolvers.
```
jpencola: change `ResourceRecordSet` to `resource record set`

src/main/java/gyro/google/dns/ZoneDnsSecConfig.java
```diff
@@ -0,0 +1,119 @@
...
+public class ZoneDnsSecConfig extends Diffable implements Copyable<ManagedZoneDnsSecConfig> {
+
+    private List<KeySpec> defaultKeySpec;
+
+    private String nonExistence;
+
+    private String state;
+
+    /**
+     * Specifies parameters for generating initial DnsKeys for this ManagedZone. Can only be changed while the state is ``off``.
```
jpencola: Change `ManagedZone` to `managed zone`.

src/main/java/gyro/google/dns/DnsPolicyNetwork.java
```diff
@@ -0,0 +1,79 @@
...
+
+        if (networkUrl != null) {
+            setNetwork(findById(NetworkResource.class, networkUrl));
+        }
+    }
+
+    @Override
+    public String primaryKey() {
+        return Optional.ofNullable(getNetwork())
+            .map(NetworkResource::primaryKey)
```
deepanjan90: Because you are using the primary key of the network resource, if the configuration is like
```
google::dns-policy dns-policy-example
    name: "dns-policy-example"
    description: "DNS Policy Example"
    network
        network: $(external-query google::compute-network {name: "vpc-example"}).0.self-link
    end
    network
        network: $(external-query google::compute-network {name: "vpc-example-asung"}).0.self-link
    end
end
```

then an error similar to this occurs after you gyro up, after create
```
$ ../../gyro/cli/dist/gyro up dns/policy.gyro --verbose
↓ Loading plugin: gyro:gyro-google-provider:0.99.1-SNAPSHOT
⟳ Refreshed resources: 1

Unexpected error: java.lang.IllegalArgumentException: Multiple entries with same key: Optional[google::compute-network::null]=network google::compute-network::null and Optional[google::compute-network::null]=network google::compute-network::null
	at com.google.common.collect.ImmutableMap.checkNoConflict(ImmutableMap.java:190)
	at com.google.common.collect.RegularImmutableMap.checkNoConflictInKeyBucket(RegularImmutableMap.java:109)
	at com.google.common.collect.RegularImmutableMap.fromEntryArray(RegularImmutableMap.java:95)
	at com.google.common.collect.ImmutableMap$Builder.build(ImmutableMap.java:373)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:582)
	at gyro.core.scope.NodeEvaluator.lambda$copy$37(NodeEvaluator.java:569)
	at java.base/java.util.Optional.map(Optional.java:265)
	at gyro.core.scope.NodeEvaluator.copy(NodeEvaluator.java:569)
	at gyro.core.scope.NodeEvaluator.lambda$visitResource$35(NodeEvaluator.java:523)
	at java.base/java.util.Optional.ifPresent(Optional.java:183)
	at gyro.core.scope.NodeEvaluator.visitResource(NodeEvaluator.java:523)
	at gyro.core.scope.NodeEvaluator.visitResource(NodeEvaluator.java:81)
	at gyro.lang.ast.block.ResourceNode.accept(ResourceNode.java:55)
	at gyro.lang.ast.NodeVisitor.visit(NodeVisitor.java:37)
	at gyro.core.scope.NodeEvaluator.lambda$evaluateBody$31(NodeEvaluator.java:341)
	at gyro.core.scope.Defer.execute(Defer.java:50)
	at gyro.core.scope.NodeEvaluator.evaluateBody(NodeEvaluator.java:341)
	at gyro.core.scope.NodeEvaluator.visitFile(NodeEvaluator.java:424)
	at gyro.core.scope.NodeEvaluator.visitFile(NodeEvaluator.java:81)
	at gyro.lang.ast.block.FileNode.accept(FileNode.java:32)
	at gyro.lang.ast.NodeVisitor.visit(NodeVisitor.java:37)
	at gyro.core.scope.NodeEvaluator.lambda$evaluateBody$31(NodeEvaluator.java:341)
	at gyro.core.scope.Defer.execute(Defer.java:50)
	at gyro.core.scope.NodeEvaluator.evaluateBody(NodeEvaluator.java:341)
	at gyro.core.scope.NodeEvaluator.evaluate(NodeEvaluator.java:325)
	at gyro.core.scope.RootScope.evaluate(RootScope.java:313)
	at gyro.core.command.AbstractConfigCommand.doExecute(AbstractConfigCommand.java:133)
	at gyro.core.command.AbstractCommand.execute(AbstractCommand.java:71)
	at gyro.cli.Gyro.run(Gyro.java:195)
	at gyro.cli.Gyro.main(Gyro.java:76)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.boot.loader.MainMethodRunner.run(MainMethodRunner.java:48)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:87)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:50)
	at org.springframework.boot.loader.JarLauncher.main(JarLauncher.java:51)
```

The reason being that the network associated is  not managed by gyro and is external, so the primary key for the Network Resource is null and there are multiple of them which creates this scenario.

Instead using the selflink as the primary key would solve this issue.

All other places where the primary key of a subresource is derived from the primary key of the parent which is a Resource should be modified to use the field which has the `@Id` field

src/main/java/gyro/google/dns/ManagedZoneResource.java
sshyub: Code style should be enforced by `checkstyle.xml` or `intellij-code-style.xml`. If it can't be done by them, a rule should be written down somewhere so that everyone can follow it.
but honestly I wouldn't be bothered if it can't be automated since there's no way we can enforce them 100%.

src/main/java/gyro/google/dns/ManagedZoneResource.java
sshyub: No, it's not allowed.

PR #68, opened 2019-12-18
Add Image

### LINKED ISSUE #61: GCP: Image
> Use operating system images to create boot disks for your instances. You can use one of the following image types:
> 
> Public images are provided and maintained by Google, open-source communities, and third-party vendors. By default, all projects have access to these images and can use them to create instances.
> 
> Custom images are available only to your project. You can create a custom image from boot disks and other images. Then, use the custom image to create an instance.

API
https://cloud.google.com/compute/docs/reference/rest/v1/images

Create Image

alex-lambrides:
Fix #61

### DIFF
--- a/src/main/java/gyro/google/compute/ImageFinder.java
+++ b/src/main/java/gyro/google/compute/ImageFinder.java
@@ -89,41 +89,39 @@ public class ImageFinder extends GoogleFinder<Compute, Image, ImageResource> {
 
     @Override
     protected List<Image> findAllGoogle(Compute client) throws Exception {
+        return listImages(client, getProjectId());
+    }
+
+    @Override
+    protected List<Image> findGoogle(Compute client, Map<String, String> filters) throws Exception {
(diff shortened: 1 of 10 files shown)


src/main/java/gyro/google/compute/ImageFinder.java
```diff
@@ -0,0 +1,129 @@
...
+            ImageList imageList;
+            String nextPageToken = null;
+
+            do {
+                imageList = client.images().list(filters.get("project")).setPageToken(nextPageToken).execute();
+                images.addAll(imageList.getItems().stream().filter(Objects::nonNull).collect(Collectors.toList()));
+                nextPageToken = imageList.getNextPageToken();
+            } while (nextPageToken != null);
+
+            return images;
```
deepanjan90: Would be a good to have a single return by declaring the list of images up top and returning it at the very ed, with the individual conditional scopes populating the list.

PR #72, opened 2019-12-19
Fix operation error exception throwing and message formatting

### LINKED ISSUE #70: GCP: waitForCompletion is always throwing unformatted exceptions
`ComputeResource#waitForCompletion` is currently always throwing an exception even in the case of a successful operation. We need to check that `operation.getErrors() != null` to ensure the operation actually errored, as currently successful operations are throwing an error like: `Error: Can't create... Caused by: Error: null`.

Also, when there is an actual error, the error message isn't formatted correctly, and looks something like this:
```
Error: Can't create google::instance-attached-disk gyro-disk-1 resource!
Caused by: Error: {
  "errors" : [ {
    "code" : "RESOURCE_IN_USE_BY_ANOTHER_RESOURCE",
    "message" : "The disk resource 'projects/aerobic-lock-236714/zones/us-west1-a/disks/instance-test-2' is already being used by 'projects/aerobic-lock-236714/zones/us-west1-a/instances/gyro-development'"
  } ]
}
```
We should update this formatting to be consistent with the rest of our error messages.

alex-lambrides:
Fixes #70

### DIFF
--- a/src/main/java/gyro/google/GoogleResource.java
+++ b/src/main/java/gyro/google/GoogleResource.java
@@ -22,6 +22,7 @@ import java.util.stream.Collectors;
 import com.google.api.client.googleapis.json.GoogleJsonError;
 import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.client.googleapis.services.json.AbstractGoogleJsonClient;
+import com.google.api.services.compute.model.Operation;
 import gyro.core.GyroException;
 import gyro.core.GyroUI;
 import gyro.core.resource.Resource;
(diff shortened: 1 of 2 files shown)


src/main/java/gyro/google/GoogleResource.java
```diff
@@ -114,6 +115,12 @@ protected static String formatGoogleExceptionMessage(GoogleJsonResponseException
             .collect(Collectors.joining("\n"));
     }
 
+    protected static String formatOperationErrorMessage(Operation operation) {
+        return operation.getError().getErrors().stream()
+            .map(Operation.Error.Errors::getMessage)
+            .collect(Collectors.joining("\n"));
+    }
+
```
deepanjan90: As this operation is part of the compute model, this should be part of `ComputeResource`.
Also better to use `Operation.Error` as the parameter, to avoid NPE if someone called this method without doing a null check

PR #76, opened 2020-01-08
Support Adding and Removing Instances to an Instance Group.

### LINKED ISSUE #75: GCP: InstanceGroupResource Add/Delete Instances
Add support for addInstances and removeInstances methods. 
https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroups

kluman:
Fixes #75

### DIFF
--- a/src/main/java/gyro/google/compute/InstanceGroupResource.java
+++ b/src/main/java/gyro/google/compute/InstanceGroupResource.java
@@ -54,6 +54,10 @@ import gyro.google.Copyable;
  *          name: "instance-group-named-ports-example"
  *          description: "instance-group-named-ports-example-description"
  *          zone: "us-central1-a"
+ *          instances: [
+ *              $(google::compute-instance gyro-instance-group-instance-a),
+ *              $(google::compute-instance gyro-instance-group-instance-b)
+ *          ]
  *
  *          named-port
  *              name: "port-a"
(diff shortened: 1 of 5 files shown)


src/main/java/gyro/google/compute/InstanceGroupResource.java
```diff
@@ -238,6 +294,55 @@ public void copyFrom(InstanceGroup instanceGroup) {
         }
     }
 
+    private void addInstances(List<String> instances) throws Exception {
+        Compute client = createComputeClient();
```
deepanjan90: The client could be a parameter, enabling it to be reused, as this method in called from `create` and `update` both creating their own clients.

src/main/java/gyro/google/compute/InstanceGroupResource.java
```diff
@@ -238,6 +294,55 @@ public void copyFrom(InstanceGroup instanceGroup) {
...
+                    .setInstances(instances.stream()
+                        .map(instance -> new InstanceReference().setInstance(instance))
+                        .collect(Collectors.toList())
+                    )
+                ).execute()
+        );
+    }
+
+    private void removeInstances(List<String> instances) throws Exception {
+        Compute client = createComputeClient();
```
deepanjan90: The client could be a parameter, enabling it to be reused, as this method in called from `update` which creates its own clients.

src/main/java/gyro/google/compute/InstanceGroupResource.java
kluman: Done.

src/main/java/gyro/google/compute/InstanceGroupResource.java
kluman: Done.

PR #87, opened 2020-01-15
Use `Wait` helper in `compute` package.

### LINKED ISSUE #86: `ComputeResource#waitForCompletion` to use `Wait` helper from `Gyro` core for consistency.
We better to implement `ComputeResource#waitForCompletion` using `gyro.core.Wait` for consistency.

sshyub:
Fixes #86

### DIFF
--- a/src/main/java/gyro/google/compute/ComputeResource.java
+++ b/src/main/java/gyro/google/compute/ComputeResource.java
@@ -16,67 +16,50 @@
 
 package gyro.google.compute;
 
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
-import com.google.api.client.googleapis.json.GoogleJsonResponseException;
 import com.google.api.services.compute.Compute;
 import com.google.api.services.compute.model.Operation;
 import gyro.core.GyroException;
+import gyro.core.Wait;
 import gyro.google.GoogleResource;
 
 public abstract class ComputeResource extends GoogleResource {
 
-    public void waitForCompletion(Compute compute, Operation operation) throws Exception {
-        waitForCompletion(compute, operation, 60000);
-    }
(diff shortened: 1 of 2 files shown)


src/main/java/gyro/google/compute/ComputeResource.java
```diff
@@ -16,67 +16,50 @@
...
-    public void waitForCompletion(Compute compute, Operation operation, long timeout) throws Exception {
-        long start = System.currentTimeMillis();
-        final long pollInterval = 1000;
-
-        String zone = operation.getZone();
-        if (zone != null) {
-            String[] bits = zone.split("/");
-            zone = bits[bits.length - 1];
-        }
+    public void waitForCompletion(Compute compute, Operation operation) {
```
deepanjan90: This method should call another method that takes in the time and time unit.
That method should implement the actual wait logic.

This way we will have option to use a custom time for the wait if needed be.

src/main/java/gyro/google/compute/ImageResource.java
```diff
@@ -451,8 +451,7 @@ protected void doCreate(GyroUI ui, State state) throws Exception {
         image.set("storageLocations", getStorageLocations());
 
         Operation operation = client.images().insert(getProjectId(), image).execute();
-        // Images are slow to complete so wait max of 3 minutes for completion
-        waitForCompletion(client, operation, 180000);
+        waitForCompletion(client, operation);
```
deepanjan90: Based on the other review, this should call the custom waitForCompletion method and specify the 3 minute wait, as this is an exceptional case. Also keep the comment as that proves this is an exceptional case.

sshyub:
No need to provide an extra API that takes time unit as requests will never be timed out.

PR #94, opened 2020-01-17
Fix NPE for all list based api calls for compute

### LINKED ISSUE #93: GCP: Fix NPE for all list based api calls for compute
A recent API change on GCP end, results in NPE when a list based api is called and there is no items present. The root cause being that the response previously returned an empty list, but now returns a null.

deepanjan90:
Fixes #93 

All places under the compute resource where a list based call is made, there is a possibility of NPE if no resource is present of that type.

Previously the api would return an empty list in the `items` if none was found, but now it returns null which causes the NPE.

### DIFF
--- a/src/main/java/gyro/google/compute/InstanceFinder.java
+++ b/src/main/java/gyro/google/compute/InstanceFinder.java
@@ -19,6 +19,7 @@ package gyro.google.compute;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import com.google.api.services.compute.Compute;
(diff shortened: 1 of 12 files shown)


src/main/java/gyro/google/compute/InstanceFinder.java
```diff
@@ -70,10 +71,10 @@ public void setFilter(String filter) {
     protected List<Instance> findAllGoogle(Compute client) throws Exception {
         List<Instance> instances = new ArrayList<>();
         String pageToken;
-        List<String> zones = client.zones().list(getProjectId()).execute().getItems()
-            .stream()
-            .map(Zone::getName)
-            .collect(Collectors.toList());
+        List<Zone> zoneList = Optional.ofNullable(client.zones().list(getProjectId()).execute().getItems()).orElse(new ArrayList<>());
```
alex-lambrides: This line exceeds the line max character count.

src/main/java/gyro/google/compute/InstanceFinder.java
```diff
@@ -70,10 +71,10 @@ public void setFilter(String filter) {
...
+        List<String> zones = zoneList.stream()
+                .map(Zone::getName)
+                .collect(Collectors.toList());
```
alex-lambrides: These should only be indented 4 rather than 8 spaces.

src/main/java/gyro/google/compute/ProjectMetadataItemFinder.java
```diff
@@ -53,20 +56,26 @@ public void setKey(String key) {
 
     @Override
     protected List<Metadata.Items> findAllGoogle(Compute client) throws Exception {
-        return client.projects().get(getProjectId()).execute().getCommonInstanceMetadata().getItems();
+        return Optional.ofNullable(client.projects().get(getProjectId()).execute().getCommonInstanceMetadata().getItems()).orElse(new ArrayList<>());
```
alex-lambrides: This line exceeds the line max character count.

PR #99, opened 2020-01-29
Target Pool and Regional Forwarding Rule

### LINKED ISSUE #79: GCP: Target Pool
> Target pools are used for network TCP/UDP load balancing.
> A Target Pool resource defines a group of instances that receive incoming traffic from forwarding rules.

This uses the target pool API: https://cloud.google.com/compute/docs/reference/rest/v1/targetPools

Target Pools

alex-lambrides:
Fixes #79.


src/main/java/gyro/google/compute/TargetPoolResource.java
```diff
@@ -0,0 +1,291 @@
...
+            Operation response = client.targetPools()
+                .setBackup(getProjectId(), getRegion(), getName(), targetReference)
+                .setFailoverRatio(getFailoverRatio())
+                .execute();
+            waitForCompletion(client, response);
+        }
+
+        if (changedFieldNames.contains("instances")) {
+            List<InstanceResource> removeInstances = currentTargetPool.getInstances();
+            removeInstances.removeAll(getInstances());
```
deepanjan90: The `removeAll()` wont work on objects unless you override the `equals` method of the class.
The objects in the collection are similar but not the same as they are instantiated separately and so are different in nature. Overriding the `equals` method gives the control back to you to compare based on defined attributes.

In the current case, the self link is a field we can look to see if two instances are common and remove it from the add/delete operation

```suggestion
            Set<String> instances = getInstances().stream()
                .map(InstanceResource::getSelfLink)
                .collect(Collectors.toSet());
            removeInstances.removeIf(o -> instances.contains(o.getSelfLink()));
```

src/main/java/gyro/google/compute/TargetPoolResource.java
```diff
@@ -0,0 +1,291 @@
...
+                        .collect(Collectors.toList()));
+                Operation response =
+                    client.targetPools()
+                        .removeInstance(getProjectId(), getRegion(), getName(), removeInstanceRequest)
+                        .execute();
+                waitForCompletion(client, response);
+            }
+
+            List<InstanceResource> newInstances = getInstances();
+            newInstances.removeAll(currentTargetPool.getInstances());
```
deepanjan90: same as above

src/main/java/gyro/google/compute/TargetPoolResource.java
```diff
@@ -0,0 +1,291 @@
...
+    public void copyFrom(TargetPool model) throws Exception {
+        setBackupPool(findById(TargetPoolResource.class, model.getBackupPool()));
+        setDescription(model.getDescription());
+        setFailoverRatio(model.getFailoverRatio());
+        setName(model.getName());
+        setSessionAffinity(model.getSessionAffinity());
+        setRegion(model.getRegion());
+        setSelfLink(model.getSelfLink());
+
+        List<String> instances = model.getInstances();
```
deepanjan90:
```suggestion
        getInstances().clear();
        List<String> instances = model.getInstances();
```

src/main/java/gyro/google/compute/ForwardingRuleResource.java
```diff
@@ -0,0 +1,127 @@
...
+        refresh();
+    }
+
+    @Override
+    public void doDelete(GyroUI ui, State state) throws Exception {
+        Compute client = createComputeClient();
+        Operation response = client.forwardingRules().delete(getProjectId(), getRegion(), getName()).execute();
+        waitForCompletion(client, response);
+    }
+}
```
deepanjan90:
```suggestion
}

```

examples/compute/forwarding-rule.gyro
```diff
@@ -0,0 +1,48 @@
...
+
+google::compute-forwarding-rule forwarding-rule-example
+    name: "forwarding-rule-example"
+    description: "Regional forwarding rule description."
+    region: "us-central1"
+    ip-protocol: "TCP"
+    load-balancing-scheme: "EXTERNAL"
+    port-range: "80-80"
+    target-pool: "https://www.googleapis.com/compute/v1/projects/aerobic-lock-236714/regions/us-central1/targetPools/test-console"
+end
```
deepanjan90:
```suggestion
end

```

PR #96, opened 2020-01-24
Regional/Zonal instance group manager.

sshyub:
Fixes #59 

Base branch must be changed to `master` once #82 is merged.

PR #100, opened 2020-01-29
Regional/Zonal Autoscaler resources.

### LINKED ISSUE #95: GCP: Autoscaler
TODO: Add link to docs and details

sshyub:
Fixes #95


sshyub:
Base branch must be changed to master once #96 is merged.

src/main/java/gyro/google/compute/AbstractAutoscalerResource.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
+    /**
+     * Server-defined URL for the resource.
+     */
+    @Output
```
deepanjan90:
```suggestion
    @Id
    @Output
```

src/main/java/gyro/google/compute/AbstractAutoscalerResource.java
```diff
@@ -0,0 +1,242 @@
...
+                autoscaler.setDescription(getDescription());
+                shouldPatch = true;
+            } else if (changedFieldName.equals("autoscaling-policy")) {
+                ComputeAutoscalingPolicy diffableAutoscalingPolicy = getAutoscalingPolicy();
+                autoscaler.setAutoscalingPolicy(diffableAutoscalingPolicy == null
+                    ? Data.nullOf(AutoscalingPolicy.class)
+                    : diffableAutoscalingPolicy.copyTo());
+                shouldPatch = true;
+            }
+        }
```
deepanjan90: Is there a specific reason to use a `for` over here ?
Having the loop checks one of the condition twice in case both the fields changes.

Can be replaced by 
```
           if (changedFieldName.contains("description")) {
                autoscaler.setDescription(getDescription());
                shouldPatch = true;
            }
           
            if (changedFieldName.contains("autoscaling-policy")) {
                ComputeAutoscalingPolicy diffableAutoscalingPolicy = getAutoscalingPolicy();
                autoscaler.setAutoscalingPolicy(diffableAutoscalingPolicy == null
                    ? Data.nullOf(AutoscalingPolicy.class)
                    : diffableAutoscalingPolicy.copyTo());
                shouldPatch = true;
            }
```

src/main/java/gyro/google/compute/AutoscalerFinder.java
```diff
@@ -0,0 +1,127 @@
...
+    protected List<Autoscaler> findGoogle(Compute client, Map<String, String> filters) throws Exception {
+        if (filters.containsKey("region")) {
+            throw new GyroException("For regional autoscaler, use 'compute-region-autoscaler' instead.");
+        }
+
+        String zone = filters.remove("zone");
+
+        if (StringUtils.isBlank(zone) || ObjectUtils.isBlank(filters)) {
+            return findAllAutoscalers(client, getProjectId(), ResourceScope.ZONE, filters);
+        }
```
deepanjan90: The find specific finder should not return everything at any point.
So in this method, `ObjectUtils.isBlank(filters)` is condition will never be true.

src/main/java/gyro/google/compute/ComputeAutoscalerStatusDetails.java
```diff
@@ -0,0 +1,102 @@
...
+        "MODE_OFF",
+        "MODE_ONLY_UP",
+        "MORE_THAN_ONE_BACKEND_SERVICE",
+        "NOT_ENOUGH_QUOTA_AVAILABLE",
+        "REGION_RESOURCE_STOCKOUT",
+        "SCALING_TARGET_DOES_NOT_EXIST",
+        "UNKNOWN",
+        "UNSUPPORTED_MAX_RATE_LOAD_BALANCING_CONFIGURATION",
+        "ZONE_RESOURCE_STOCKOUT"
+    })
```
deepanjan90: As this is an output field, we do not need validation on it.

src/main/java/gyro/google/compute/ComputeAutoscalingPolicy.java
```diff
@@ -0,0 +1,238 @@
...
+
+    private Integer maxNumReplicas;
+
+    private Integer minNumReplicas;
+
+    private String mode;
+
+    /**
+     * The number of seconds that the autoscaler should wait before it starts collecting information from a new instance.
+     * This prevents the autoscaler from collecting information when the instance is initializing, during which the collected usage would not be reliable. The default time autoscaler waits is 60 seconds. Virtual machine initialization times might vary because of numerous factors. We recommend that you test how long an instance may take to initialize. To do this, create an instance and time the startup process.
```
deepanjan90:
```suggestion
     * This prevents the autoscaler from collecting information when the instance is initializing, during which the collected usage would not be reliable. The default time autoscaler waits is ``60`` seconds. Virtual machine initialization times might vary because of numerous factors. We recommend that you test how long an instance may take to initialize. To do this, create an instance and time the startup process.
```

src/main/java/gyro/google/compute/ComputeAutoscalingPolicy.java
```diff
@@ -0,0 +1,238 @@
...
+    public Integer getMinNumReplicas() {
+        return minNumReplicas;
+    }
+
+    public void setMinNumReplicas(Integer minNumReplicas) {
+        this.minNumReplicas = minNumReplicas;
+    }
+
+    /**
+     * Operating mode for this policy.
```
deepanjan90:
```suggestion
     * Operating mode for this policy. Valid values are ``OFF``, ``ON`` or ``ONLY_UP``.
```

src/main/java/gyro/google/compute/ComputeAutoscalingPolicyCustomMetricUtilization.java
```diff
@@ -0,0 +1,103 @@
...
+        this.metric = metric;
+    }
+
+    /**
+     * The target value of the metric that autoscaler should maintain.
+     * This must be a positive value. A utilization metric scales number of virtual machines handling requests to increase or decrease proportionally to the metric. For example, a good metric to use as a utilization_target is compute.googleapis.com/instance/network/received_bytes_count. The autoscaler will work to keep this value constant for each of the instances.
+     */
+    @Required
+    @Updatable
+    public Double getUtilizationTarget() {
```
deepanjan90: As the doc says needs be positive.

```suggestion
    @Min(0)
    public Double getUtilizationTarget() {
```

src/main/java/gyro/google/compute/RegionAutoscalerFinder.java
```diff
@@ -0,0 +1,86 @@
...
+    protected List<Autoscaler> findGoogle(Compute client, Map<String, String> filters) throws Exception {
+        if (filters.containsKey("zone")) {
+            throw new GyroException("For zonal autoscaler, use 'compute-autoscaler' instead.");
+        }
+
+        String region = filters.remove("region");
+
+        if (StringUtils.isBlank(region) || ObjectUtils.isBlank(filters)) {
+            return AutoscalerFinder.findAllAutoscalers(client, getProjectId(), ResourceScope.REGION, filters);
+        }
```
deepanjan90: The find specific finder should not return everything at any point. I would argue that it should throw an exception if region is blank.

src/main/java/gyro/google/compute/ComputeAutoscalingPolicyLoadBalancingUtilization.java
```diff
@@ -0,0 +1,61 @@
...
+import gyro.google.Copyable;
+
+public class ComputeAutoscalingPolicyLoadBalancingUtilization extends Diffable
+    implements Copyable<AutoscalingPolicyLoadBalancingUtilization> {
+
+    private Double utilizationTarget;
+
+    /**
+     * Fraction of backend capacity utilization (set in HTTP(S) load balancing configuration) that autoscaler should maintain.
+     * Must be a positive float value. If not defined, the default is 0.8.
```
deepanjan90:
```suggestion
     * Must be a positive float value. If not defined, the default is ``0.8``.
```

src/main/java/gyro/google/compute/ComputeAutoscalingPolicyCpuUtilization.java
```diff
@@ -0,0 +1,61 @@
...
+import gyro.google.Copyable;
+
+public class ComputeAutoscalingPolicyCpuUtilization extends Diffable
+    implements Copyable<AutoscalingPolicyCpuUtilization> {
+
+    private Double utilizationTarget;
+
+    /**
+     * The target CPU utilization that the autoscaler should maintain.
+     * Must be a float value in the range (0, 1]. If not specified, the default is 0.6. If the CPU level is below the target utilization, the autoscaler scales down the number of instances until it reaches the minimum number of instances you specified or until the average CPU of your instances reaches the target utilization. If the average CPU is above the target utilization, the autoscaler scales up until it reaches the maximum number of instances you specified or until the average utilization reaches the target utilization.
```
deepanjan90:
```suggestion
     * Must be a float value in the range ``(0, 1]``. If not specified, the default is ``0.6``. If the CPU level is below the target utilization, the autoscaler scales down the number of instances until it reaches the minimum number of instances you specified or until the average CPU of your instances reaches the target utilization. If the average CPU is above the target utilization, the autoscaler scales up until it reaches the maximum number of instances you specified or until the average utilization reaches the target utilization.
```

src/main/java/gyro/google/compute/ComputeAutoscalingPolicy.java
```diff
@@ -0,0 +1,238 @@
...
+        return maxNumReplicas;
+    }
+
+    public void setMaxNumReplicas(Integer maxNumReplicas) {
+        this.maxNumReplicas = maxNumReplicas;
+    }
+
+    /**
+     * The minimum number of replicas that the autoscaler can scale down to.
+     * This cannot be less than 0. If not provided, autoscaler will choose a default value depending on maximum number of instances allowed.
```
deepanjan90:
```suggestion
     * This cannot be less than ``0``. If not provided, autoscaler will choose a default value depending on maximum number of instances allowed.
```

src/main/java/gyro/google/compute/AbstractAutoscalerResource.java
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * The configuration parameters for the autoscaling algorithm.
+     * You can define one or more of the policies for an autoscaler:
+     *
+     * - ``cpu-utilization``
+     * - ``custom-metric-utilizations``
+     * - ``load-balancing-utilization``
+     *
+     * If none of these are specified, the default will be to autoscale based on cpuUtilization to 0.6 or 60%.
```
deepanjan90:
```suggestion
     * If none of these are specified, the default will be to autoscale based on cpuUtilization to ``0.6`` or `` 60%``.
```

src/main/java/gyro/google/compute/AbstractAutoscalerResource.java
sshyub: There's no obviously better solution as the suggested solution checks both conditions in case only one field changes, so best solution depends on the actual use cases.
Fixes have been provided even though I still think it's too much.

src/main/java/gyro/google/compute/AutoscalerFinder.java
sshyub: it doesn't return everything, but still uses the filters a user provided, and what it returns is totally  up to the provider.

src/main/java/gyro/google/compute/RegionAutoscalerFinder.java
sshyub: it doesn't return everything, but still uses the filters a user provided, and what it returns is totally  up to the provider.

PR #101, opened 2020-01-30
Bucket IAM Policy implementation

### LINKED ISSUE #63: GCP: Bucket IAM Policy
Updates an IAM policy for a specified bucket.

https://cloud.google.com/storage/docs/json_api/v1/buckets/setIamPolicy

jpencola:
Fixes #63

### DIFF
--- a/src/main/java/gyro/google/storage/BucketIamPolicy.java
+++ b/src/main/java/gyro/google/storage/BucketIamPolicy.java
@@ -67,11 +67,11 @@ public class BucketIamPolicy extends Diffable implements Copyable<Policy> {
     }
 
     /**
-     * The version of this policy. This is always set to ``3``. See also `Understanding Policies <https://cloud.google.com/iam/docs/policies#versions>`_.
+     * The version of this policy. See also `Understanding Policies <https://cloud.google.com/iam/docs/policies#versions>`_.
      */
     @Output
     public Integer getVersion() {
-        return 3;
+        return version;
     }
 
     public void setVersion(Integer version) {
@@ -85,7 +85,7 @@ public class BucketIamPolicy extends Diffable implements Copyable<Policy> {
 
     @Override
     public void copyFrom(Policy model) {
-        setVersion(3);
+        setVersion(model.getVersion());
         setResourceId(model.getResourceId());
         getBindings().clear();
         if (model.getBindings() != null) {
@@ -99,7 +99,7 @@ public class BucketIamPolicy extends Diffable implements Copyable<Policy> {
     }
 
     public Policy toPolicy() {
-        return new Policy().setVersion(getVersion()).setResourceId(getResourceId())
+        return new Policy().setVersion(3).setResourceId(getResourceId())
             .setBindings(getBindings().stream().map(BucketIamPolicyBinding::toBinding).collect(Collectors.toList()));
     }
 }
--- a/src/main/java/gyro/google/storage/BucketIamPolicy.java
+++ b/src/main/java/gyro/google/storage/BucketIamPolicy.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2020, Perfect Sense, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


src/main/java/gyro/google/storage/BucketIamPolicy.java
```diff
@@ -0,0 +1,105 @@
...
+        this.resourceId = resourceId;
+    }
+
+    /**
+     * The version of this policy. This is always set to ``3``. See also `Understanding Policies <https://cloud.google.com/iam/docs/policies#versions>`_.
+     */
+    @Output
+    public Integer getVersion() {
+        return 3;
+    }
```
deepanjan90: Because the version is not exposed as a field in the the console and you are not really using this field, it could be just stated as a version and maybe what it means, but the copy from should set it to what ever version the policy returns.

src/main/java/gyro/google/storage/BucketIamPolicyBinding.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public void setCondition(BucketIamPolicyBindingCondition condition) {
+        this.condition = condition;
+    }
+
+    @Override
+    public String primaryKey() {
+        return String.format(
+            "where role is \"%s\" and members are \"%s\" with condition \"%s\"",
```
deepanjan90:
```suggestion
            "with role  '%s', members [ %s ] with condition '%s',
```

src/main/java/gyro/google/storage/BucketIamPolicyBinding.java
```diff
@@ -0,0 +1,116 @@
...
+            getRole(),
+            String.join(",", getMembers()),
```
deepanjan90:
```suggestion
            String.join("','", getMembers()),
```

src/main/java/gyro/google/storage/BucketIamPolicyBinding.java
```diff
@@ -0,0 +1,116 @@
...
+            (getCondition() == null) ? "" : getCondition());
```
deepanjan90:
```suggestion
            (getCondition() == null) ? "" : getCondition().primarykey());
```

src/main/java/gyro/google/storage/BucketIamPolicyBindingCondition.java
```diff
@@ -0,0 +1,85 @@
...
+        return title;
+    }
+
+    public void setTitle(String title) {
+        this.title = title;
+    }
+
+    @Override
+    public String primaryKey() {
+        return String.format("with title \"%s\" and expression of \"%s\"", getTitle(), getExpression());
```
deepanjan90:
```suggestion
        return String.format("with title '%s' and expression of '%s'", getTitle(), getExpression());
```

src/main/java/gyro/google/storage/BucketResource.java
```diff
@@ -77,6 +78,27 @@
  *              end
  *          end
  *
+ *          iam-policy
+ *              bindings
+ *                  role: 'roles/storage.legacyBucketOwner'
+ *                  members: ['projectEditor:<<PROJECT-ID-GOES-HERE>>']
```
deepanjan90:
```suggestion
 *                  members: ['projectEditor:<<PROJECT-ID-GOES-HERE>>'/the name of the role]
```

src/main/java/gyro/google/storage/BucketResource.java
```diff
@@ -77,6 +78,27 @@
...
+ *              end
+ *
+ *              bindings
+ *                  role: 'roles/storage.legacyBucketReader'
+ *                  members: ['projectViewer:<<PROJECT-ID-GOES-HERE>>']
```
deepanjan90: same as above

src/main/java/gyro/google/storage/BucketResource.java
```diff
@@ -552,7 +605,20 @@ public void doDelete(GyroUI ui, State state) throws Exception {
...
+        Policy iamPolicy = storage.buckets()
+            .getIamPolicy(getName())
+            .set("optionsRequestedPolicyVersion", 3)
+            .execute();
+        if (iamPolicy != null) {
+            BucketIamPolicy bucketIamPolicy = newSubresource(BucketIamPolicy.class);
+            bucketIamPolicy.copyFrom(iamPolicy);
+            setIamPolicy(bucketIamPolicy);
+        }
+
```
deepanjan90: Move this to the end of the copyFrom()
1. As it uses the `name` field to fetch the iam policy, So if called from an external query, the `name` field would not be set and will be null and cause an error.
2. when adding things to an existing method it is a good practice to add things at the end unless there is need to add it before existing code.

src/main/java/gyro/google/storage/BucketResource.java
jpencola: When using `projectEditor`, the member value should just be the project id.

PR #130, opened 2020-03-19
KMS Key Rings and Crypto Keys

### LINKED ISSUE #116: GCP: KMS Key Rings
>A KeyRing is a toplevel logical grouping of CryptoKeys.

This uses the KMS Key Rings API:
https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings

Screen Shot 2020-03-10 at 2 55 48 PM

Related to #117

### LINKED ISSUE #117: GCP: KMS Crypto Keys
>A CryptoKey represents a logical key that can be used for cryptographic operations.

>A CryptoKey is made up of one or more versions, which represent the actual key material used in cryptographic operations.

This uses the KMS Key Rings Crypto Keys API:
https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys

Screen Shot 2020-03-10 at 2 58 37 PM

Related to #116

harjain99:
Fixes #116 and Fixes #117


src/main/java/gyro/google/kms/CryptoKeyFinder.java
```diff
@@ -0,0 +1,73 @@
...
+ * -------
+ *
+ * .. code-block:: gyro
+ *
+ *    crypto-key: $(external-query google::crypto-key {location: "us-east4", key-ring-name: "key-ring-example"})
+ */
+@Type("crypto-key")
+public class CryptoKeyFinder extends GoogleFinder<KeyManagementServiceClient, CryptoKey, CryptoKeyResource> {
+    private String location;
+    private String keyRingName;
```
deepanjan90: rename variable and corresponding use cases and example
```suggestion
    private String name;
```

src/main/java/gyro/google/kms/CryptoKeyFinder.java
```diff
@@ -0,0 +1,73 @@
...
+        return keyRingName;
+    }
+
+    public void setKeyRingName(String keyRingName) {
+        this.keyRingName = keyRingName;
+    }
+
+    @Override
+    protected List<CryptoKey> findAllGoogle(KeyManagementServiceClient client) throws Exception {
+        throw new GyroException("'location' and 'key-ring-name' are required filters");
```
deepanjan90: why are you throwing a 
```suggestion
        throw new UnsupportedOperationException("Finding all `crypto-keys` without any filter is not supported!!");
```

src/main/java/gyro/google/kms/CryptoKeyFinder.java
```diff
@@ -0,0 +1,73 @@
...
+            KeyManagementServiceClient.ListCryptoKeysPagedResponse response = client.listCryptoKeys(KeyRingName.format(
+                getProjectId(),
+                filters.get("location"),
+                filters.get("key-ring-name")));
+            response.iterateAll().forEach(keys::add);
+
+            return keys;
+        } else {
+            throw new GyroException("'location' and 'key-ring-name' are required filters");
+        }
```
deepanjan90: By convention we do not check for filter content

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,382 @@
...
+ *        end
+ *    end
+ */
+@Type("crypto-key")
+public class CryptoKeyResource extends GoogleResource implements Copyable<CryptoKey> {
+
+    private KeyRingResource keyRing;
+    private String name;
+    private Long rotationPeriod;
+    private CryptoKey.CryptoKeyPurpose purpose;
```
deepanjan90: If `import com.google.cloud.kms.v1.CryptoKey;` is included
```suggestion
    private CryptoKeyPurpose purpose;
```

src/main/java/gyro/google/kms/CryptoKeyFinder.java
harjain99: The api that fetches the resources requires the `location` and `key-ring-name` fields. Should we return an empty list instead if either of the fields aren't provided?

src/main/java/gyro/google/kms/CryptoKeyVersionFinder.java
```diff
@@ -0,0 +1,88 @@
...
+        return keyName;
+    }
+
+    public void setKeyName(String keyName) {
+        this.keyName = keyName;
+    }
+
+    @Override
+    protected List<CryptoKeyVersion> findAllGoogle(KeyManagementServiceClient client) throws Exception {
+        throw new GyroException("'location', 'key-ring-name' and 'key-name' are required filters");
```
deepanjan90:
```suggestion
        throw new UnsupportedOperationException("Finding all `crypto-key-versions` without any filter is not supported!!");
```

src/main/java/gyro/google/kms/CryptoKeyVersionFinder.java
```diff
@@ -0,0 +1,88 @@
...
+                    getProjectId(),
+                    filters.get("location"),
+                    filters.get("key-ring-name"),
+                    filters.get("key-name")));
+            response.iterateAll().forEach(keys::add);
+
+            return keys;
+        } else {
+            throw new GyroException("'location', 'key-ring-name' and 'key-name' are required filters");
+        }
```
deepanjan90: By convention we don't check for filters

src/main/java/gyro/google/kms/CryptoKeyVersionResource.java
```diff
@@ -0,0 +1,143 @@
...
+ *    google::crypto-key-version crypto-key-version-example
+ *        crypto-key: $(google::crypto-key example-crypto-key)
+ *        state: ENABLED
+ *    end
+ */
+@Type("crypto-key-version")
+public class CryptoKeyVersionResource extends GoogleResource implements Copyable<CryptoKeyVersion> {
+
+    private CryptoKeyResource cryptoKey;
+    private CryptoKeyVersion.CryptoKeyVersionState state;
```
deepanjan90: If imported `import com.google.cloud.kms.v1.CryptoKeyVersion;`
```suggestion
    private CryptoKeyVersionState state;
```

src/main/java/gyro/google/kms/CryptoKeyVersionResource.java
```diff
@@ -0,0 +1,143 @@
...
+
+    // Read-Only
```
deepanjan90:
```suggestion
    // Read-only
```

src/main/java/gyro/google/kms/CryptoKeyVersionResource.java
```diff
@@ -0,0 +1,143 @@
...
+    public CryptoKeyResource getCryptoKey() {
+        return cryptoKey;
+    }
+
+    public void setCryptoKey(CryptoKeyResource cryptoKey) {
+        this.cryptoKey = cryptoKey;
+    }
+
+    /**
+     * The state of the crypto key version. The default value is ``ENABLED``.
```
deepanjan90:
```suggestion
     * The state of the crypto key version. Defaults to ``ENABLED``.
```

src/main/java/gyro/google/kms/CryptoKeyVersionResource.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public void setState(CryptoKeyVersion.CryptoKeyVersionState state) {
+        this.state = state;
+    }
+
+    /**
+     * The ID of the crypto key version.
+     */
+    @Id
```
deepanjan90:
```suggestion
    @Id
    @Output
```

src/main/java/gyro/google/kms/CryptoKeyVersionTemplate.java
```diff
@@ -0,0 +1,64 @@
...
+            .setAlgorithm(getAlgorithm())
+            .setProtectionLevel(getProtectionLevel())
+            .build();
+    }
+
+    com.google.cloud.kms.v1.CryptoKeyVersionTemplate updateCryptoKeyVersionAlgorithm() {
+        return com.google.cloud.kms.v1.CryptoKeyVersionTemplate.newBuilder()
+            .setAlgorithm(getAlgorithm())
+            .build();
+    }
```
deepanjan90: why not use `toCryptoKeyVersionTemplate` ?

src/main/java/gyro/google/kms/KeyRingFinder.java
```diff
@@ -0,0 +1,95 @@
...
+            "us-west1",
+            "us-west2",
+            "us-west3",
+            "southamerica-east1",
+            "eur4",
+            "nam4",
+            "global",
+            "asia",
+            "europe",
+            "us");
```
deepanjan90: Figure out if there is an api to get the locations

src/main/java/gyro/google/kms/KeyRingFinder.java
```diff
@@ -0,0 +1,95 @@
...
+            List<KeyRing> keyRings = new ArrayList<>();
+
+            KeyManagementServiceClient.ListKeyRingsPagedResponse response = client.listKeyRings(
+                LocationName.format(getProjectId(), filters.get("location")));
+            response.iterateAll().forEach(k -> keyRings.add(k));
+
+            return keyRings;
+        } else {
+            throw new GyroException("'location' is a required filter");
+        }
```
deepanjan90: by convention we don not check filters

src/main/java/gyro/google/kms/KeyRingResource.java
```diff
@@ -0,0 +1,164 @@
...
+        "us-west2",
+        "us-west3",
+        "southamerica-east1",
+        "eur4",
+        "nam4",
+        "global",
+        "asia",
+        "europe",
+        "us"
+    })
```
deepanjan90: Not the best idea, because if they update their region, it breaks this

src/main/java/gyro/google/kms/KeyRingResource.java
```diff
@@ -0,0 +1,164 @@
...
+
+        return true;
+    }
+
+    @Override
+    protected void doCreate(GyroUI ui, State state) throws Exception {
+        KeyManagementServiceClient client = createClient(KeyManagementServiceClient.class);
+
+        String parent = LocationName.format(getProjectId(), getLocation());
+        KeyRing response = client.createKeyRing(parent, getName(), KeyRing.newBuilder().build());
```
deepanjan90: Makes the code more readable, parent here is a bit ambiguous
```suggestion
        String location = LocationName.format(getProjectId(), getLocation());
        KeyRing response = client.createKeyRing(location, getName(), KeyRing.newBuilder().build());
```

src/main/java/gyro/google/kms/KeyRingFinder.java
harjain99: Already looked into it and couldn't find any options. We had a discussion too if I remember correctly.

src/main/java/gyro/google/kms/KeyRingResource.java
```diff
@@ -0,0 +1,164 @@
...
+
+    }
+
+    @Override
+    protected void doDelete(GyroUI ui, State state) throws Exception {
+
+    }
+
+    public String getLocationFromId() {
+        return getId().split("/")[3];
```
deepanjan90: Do you think splitting based on a string is better than `/` as if they prepend something to the url this will break ?

src/main/java/gyro/google/kms/CryptoKeyVersionTemplate.java
harjain99: While updating, setting the protection level throws an error since its not configurable.

src/main/java/gyro/google/kms/KeyRingResource.java
harjain99: There is no api to get a list of the regions. My alternative solution would be to throw an UnexpectedOperationException if the `location` is not supplied.

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,382 @@
...
+
+        if (model.getVersionTemplate() != null) {
+            CryptoKeyVersionTemplate cryptoKeyVersionTemplate = newSubresource(CryptoKeyVersionTemplate.class);
+            cryptoKeyVersionTemplate.copyFrom(model.getVersionTemplate());
+            setCryptoKeyVersionTemplate(cryptoKeyVersionTemplate);
+        }
+
+        if (model.hasPrimary()) {
+            setPrimaryKeyVersionId(model.getPrimary().getName().split("/")[9]);
+        }
```
deepanjan90: else clause ???

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,382 @@
...
+    public String getNextRotationDate() {
+        return nextRotationDate;
+    }
+
+    public void setNextRotationDate(String nextRotationDate) {
+        this.nextRotationDate = nextRotationDate;
+    }
+
+    /**
+     * The period after which the symmetric key should automatically rotate. Minimum value ``1``.
```
deepanjan90:
```suggestion
     * The period after which the symmetric key should automatically rotate. Minimum value is ``1``.
```

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,382 @@
...
+
+    public void setKeyRing(KeyRingResource keyRing) {
+        this.keyRing = keyRing;
+    }
+
+    /**
+     * The name of the crypto key. (Required)
+     */
+    @Required
+    @Regex("^([a-z]|[0-9]|-|_)*$")
```
deepanjan90:
```suggestion
    @Regex("^([a-z]|[0-9]|-|_)*$", "message")
```

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,382 @@
...
+        KeyManagementServiceClient client = createClient(KeyManagementServiceClient.class);
+
+        CryptoKey cryptoKey = client.getCryptoKey(getId());
+
+        if (cryptoKey == null) {
+            return false;
+        }
+
+        copyFrom(cryptoKey);
+        refreshVersions(client);
```
deepanjan90: why is `refreshVersions` not part of copyFrom() ??

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,382 @@
...
+        GyroUI ui, State state, Resource current, Set<String> changedFieldNames) throws Exception {
+        KeyManagementServiceClient client = createClient(KeyManagementServiceClient.class);
+
+        FieldMask.Builder fieldMaskBuilder = FieldMask.newBuilder();
+        CryptoKey.Builder cryptoKeyBuilder = CryptoKey.newBuilder().setName(getId());
+
+        if (changedFieldNames.contains("rotation-period")) {
+            cryptoKeyBuilder.setRotationPeriod(Duration.newBuilder().setSeconds(getRotationPeriod() * 86400).build());
+            fieldMaskBuilder.addPaths("rotation_period");
+        }
```
deepanjan90: NPE if `getRotationPeriod ` is set to null.

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,382 @@
...
+
+        if (changedFieldNames.contains("crypto-key-version-template")) {
+            cryptoKeyBuilder.setVersionTemplate(getCryptoKeyVersionTemplate().updateCryptoKeyVersionAlgorithm());
+            fieldMaskBuilder.addPaths("version_template.algorithm");
+        }
```
deepanjan90: NPE if `getCryptoKeyVersionTemplate` is set to null.

src/main/java/gyro/google/kms/CryptoKeyResource.java
harjain99: I can remove the null check for the `getVersionTemplate` since its s required field and it can't be null. I don't exactly remember why I added the null check, but I'll keep you updated in case I can't.

For the others, I don't think we require else clauses since the fields are null values anyways.

src/main/java/gyro/google/kms/CryptoKeyResource.java
harjain99: From the method declaration, I thought `copyFrom` takes values from a given model, while `refreshVersions` needs access to the client to get another resource.

src/main/java/gyro/google/kms/CryptoKeyFinder.java
deepanjan90: assume they are there, we typically dont validate on finders

src/main/java/gyro/google/kms/CryptoKeyVersionTemplate.java
deepanjan90: even if you set it to something that is already set ?

src/main/java/gyro/google/kms/KeyRingFinder.java
deepanjan90: if no api is there to list locations then we should not validate this

src/main/java/gyro/google/kms/KeyRingResource.java
deepanjan90: wouldn't the api throw one, if supplied with a wrong string

src/main/java/gyro/google/kms/CryptoKeyResource.java
deepanjan90: copyFrom sets values based on the model, if that requires further api calls then the copyFrom does that as well, you just create a fresh client inside the copyFrom

src/main/java/gyro/google/kms/KeyRingResource.java
harjain99: According to the api, the ids of these resources are of the form:
`projects/*/locations/*/keyRings/*` (for the keyRing)
`projects/*/locations/*/keyRings/*/cryptoKeys/` (for the key)
`projects/*/locations/*/keyRings/*/cryptoKeys/*/cryptoKeyVersions/*` (for the version).
I doubt this would be different, but to be on a safer side, we could do something like:
`
List<String> list = Arrays.asList(getId().split("/"));
int index = list.indexOf("locations");
return list.get(index+1);
`
What do you think?

src/main/java/gyro/google/kms/CryptoKeyVersionTemplate.java
harjain99: Yep

src/main/java/gyro/google/kms/KeyRingResource.java
harjain99: Nope it doesn't. If an invalid string is provided, the api sets it to `global`.

src/main/java/gyro/google/kms/KeyRingFinder.java
harjain99: It's not a validation in this case. For the `findAllGoogle` method, I'm getting a list of keyRings from all locations.

deepanjan90: Given the nature of this, I would say skip the list and throw an unsupported exception for this method then

src/main/java/gyro/google/kms/KeyRingResource.java
deepanjan90: That sounds reasonable, see if you can make a utility method as there is a number of places you are splitting the id to get these kind of info

src/main/java/gyro/google/kms/CryptoKeyResource.java
harjain99: This field can't be changed to null. It's a required based on the `purpose` which is not updatable.

deepanjan90: Is there a validation somewhere for that ?

harjain99: Yep, the `validate()` method does it.

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,372 @@
...
+
+        return errors;
+    }
+
+    private void refreshVersions() {
+        KeyManagementServiceClient client = createClient(KeyManagementServiceClient.class);
+        getVersions().clear();
+        client.listCryptoKeyVersions(getId())
+            .iterateAll()
+            .forEach(v -> getVersions().add(v.getName()));
```
deepanjan90:
```suggestion
            .forEach(v -> getVersions().add(v.getName()));
            
            client.shutDown()
```

src/main/java/gyro/google/kms/KeyRingResource.java
```diff
@@ -0,0 +1,140 @@
...
+public class KeyRingResource extends GoogleResource implements Copyable<KeyRing> {
+
+    private String location;
+    private String name;
+
+    // Read-only
+    private String id;
+
+    /**
+     * The location of the key ring. The valid values are ``asia-east1`` or ``asia-east2`` or ``asia-northeast1`` or ``asia-northeast2`` or ``asia-northeast3`` or ``asia-south1`` or ``asia-southeast1`` or ``australia-southeast1`` or ``europe-north1`` or ``europe-west1`` or ``europe-west2`` or ``europe-west3`` or ``europe-west4`` or ``europe-west6`` or ``northamerica-northeast1`` or ``us-central1`` or ``us-east1`` or ``us-east4`` or ``us-west1`` or ``us-west2`` or ``us-west3`` or ``southamerica-east1`` or ``eur4`` or ``nam4`` or ``global`` or ``asia`` or ``europe`` or ``us``. (Required)
```
deepanjan90: I would say remove the valid values from the docs as well, as even that needs to be updated when a value changes

src/main/java/gyro/google/kms/KeyRingResource.java
```diff
@@ -0,0 +1,140 @@
...
+
+    public void setLocation(String location) {
+        this.location = location;
+    }
+
+    /**
+     * The name of the key ring. Can be letters, numbers, underscores or hyphens. (Required)
+     */
+    @Required
+    @Regex("^([a-z]|[0-9]|-|_)*$")
```
deepanjan90: give an appropriate message in the regex

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,372 @@
...
+
+    public void setVersions(List<String> versions) {
+        this.versions = versions;
+    }
+
+    @Override
+    public void copyFrom(CryptoKey model) throws Exception {
+        setId(model.getName());
+        setPurpose(model.getPurpose());
+        setKeyRing(findById(KeyRingResource.class, String.join("/", Arrays.copyOfRange(getId().split("/"), 0, 6))));
```
deepanjan90: should this be using one of the utils ?

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        if (model.getVersionTemplate() != null) {
+            CryptoKeyVersionTemplate cryptoKeyVersionTemplate = newSubresource(CryptoKeyVersionTemplate.class);
+            cryptoKeyVersionTemplate.copyFrom(model.getVersionTemplate());
+            setCryptoKeyVersionTemplate(cryptoKeyVersionTemplate);
+        }
+
+        if (model.hasPrimary()) {
+            setPrimaryKeyVersionId(model.getPrimary().getName().split("/")[9]);
```
deepanjan90: should this be using the utils ?

src/main/java/gyro/google/kms/KeyRingResource.java
```diff
@@ -0,0 +1,156 @@
...
+
+    public void setLocation(String location) {
+        this.location = location;
+    }
+
+    /**
+     * The name of the key ring. Can be letters, numbers, underscores or hyphens. (Required)
+     */
+    @Required
+    @Regex(value = "^(\\w|-|_)+$", message = "The name can be a string containing letters number")
```
deepanjan90: Doc and regex message dont match

src/main/java/gyro/google/kms/KeyRingResource.java
```diff
@@ -0,0 +1,156 @@
...
+    protected void doCreate(GyroUI ui, State state) throws Exception {
+        KeyManagementServiceClient client = createClient(KeyManagementServiceClient.class);
+
+        try {
+            client.listKeyRings(LocationName.format(getProjectId(), getLocation()));
+
+        } catch (NotFoundException ex) {
+            if (ex.getStatusCode().getCode().equals(StatusCode.Code.NOT_FOUND)) {
+                throw new GyroException(String.format(
+                    "The value provided for the location, %s, is invalid. Please provide a valid location.",
```
deepanjan90:
```suggestion
                    "Invalid value: %s for 'location'.",
```

src/main/java/gyro/google/kms/CryptoKeyResource.java
```diff
@@ -0,0 +1,389 @@
...
+public class CryptoKeyResource extends GoogleResource implements Copyable<CryptoKey> {
+
+    private KeyRingResource keyRing;
+    private String name;
+    private Long rotationPeriod;
+    private CryptoKeyPurpose purpose;
+    private String nextRotationDate;
+    private CryptoKeyVersionTemplate cryptoKeyVersionTemplate;
+    private Map<String, String> labels;
+    private String primaryKeyVersionId;
```
deepanjan90: I feel this should be `CryptoKeyVersionResource` as we have that implemented,
also from the console the resource id you get is the full id and not just the number as you are expecting

PR #135, opened 2020-03-24
SSL Certificates

### LINKED ISSUE #115: GCP: SSL Certificates
>Represents an SSL Certificate resource.
>The sslCertificates are used by:

>external HTTPS load balancers
>SSL proxy load balancers
>The regionSslCertificates are used by internal HTTPS load balancers.

>This SSL certificate resource also contains a private key. You can use SSL keys and certificates to secure connections to a load balancer. For more information, read Creating and Using SSL Certificates.

This uses the SSL Ceritifactes:
https://cloud.google.com/compute/docs/reference/rest/v1/sslCertificates

This would should include the region based as well:
https://cloud.google.com/compute/docs/reference/rest/v1/regionSslCertificates

It may be impossible to create a secure certificate since the private key needs to be given. If this is the case, support should still be extended to Google managed certs applied to HTTPS Target Proxies.

Related to Related to #114

alex-lambrides:
Fixes #115

### DIFF
--- a/src/main/java/gyro/google/compute/AbstractSslCertificateResource.java
+++ b/src/main/java/gyro/google/compute/AbstractSslCertificateResource.java
@@ -16,16 +16,18 @@
 
 package gyro.google.compute;
 
-import java.nio.file.Files;
-import java.nio.file.Paths;
+import java.io.InputStreamReader;
 
 import com.google.api.services.compute.model.SslCertificate;
+import com.google.common.io.CharStreams;
 import gyro.core.GyroException;
 import gyro.core.resource.Id;
 import gyro.core.resource.Output;
 import gyro.core.validation.Regex;
 import gyro.core.validation.Required;
 import gyro.google.Copyable;
+
+import static java.nio.charset.StandardCharsets.*;
 
 public abstract class AbstractSslCertificateResource extends ComputeResource implements Copyable<SslCertificate> {
 
@@ -38,7 +40,7 @@ public abstract class AbstractSslCertificateResource extends ComputeResource imp
     private String selfLink;
 
     /**
-     * The full path to the file containing the certificate(s) (in PEM format). The certificate chain must be no greater than 5 certs long. The chain must include at least one intermediate cert. Only used when creating the SSL certificate. (Required)
+     * The full path to the file containing the certificate(s) (in PEM format). The certificate chain must be no greater than 5 certs long. The chain must include at least one intermediate cert. (Required)
      */
     @Required
     public String getCertificatePath() {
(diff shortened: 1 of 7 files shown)


src/main/java/gyro/google/compute/AbstractSslCertificateResource.java
```diff
@@ -0,0 +1,122 @@
...
+    private String certificatePath;
+    private String description;
+    private String name;
+    private String privateKeyPath;
+
+    // Read-only
+    private String selfLink;
+
+    /**
+     * The full path to the file containing the certificate(s) (in PEM format). The certificate chain must be no greater than 5 certs long. The chain must include at least one intermediate cert. Only used when creating the SSL certificate. (Required)
```
deepanjan90: Redundant as update is not supported.
```suggestion
     * The full path to the file containing the certificate(s) (in PEM format). The certificate chain must be no greater than 5 certs long. The chain must include at least one intermediate cert. (Required)
```

src/main/java/gyro/google/compute/AbstractSslCertificateResource.java
```diff
@@ -0,0 +1,122 @@
...
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    /**
+     * The full path to the file containing the private key (in PEM format). Only used when creating the SSL certificate. (Required)
```
deepanjan90: Redundant as update is not supported.
```suggestion
     * The full path to the file containing the private key (in PEM format). (Required)
```

src/main/java/gyro/google/compute/AbstractSslCertificateResource.java
```diff
@@ -0,0 +1,122 @@
...
+    @Override
+    public void copyFrom(SslCertificate model) throws Exception {
+        setDescription(model.getDescription());
+        setName(model.getName());
+        setSelfLink(model.getSelfLink());
+    }
+
+    protected String readCertificateFile() {
+        try {
+            return new String(Files.readAllBytes(Paths.get(getCertificatePath())));
```
deepanjan90: The `openInput()` utility method helps resolve relative path
So in the configs rather than providing the full path relative to the configuration the path could be set
```suggestion
           return IoUtils.toString(openInput(getCertificatePath()), UTF_8);
```

src/main/java/gyro/google/compute/AbstractSslCertificateResource.java
```diff
@@ -0,0 +1,122 @@
...
+        } catch (Exception ex) {
+            throw new GyroException("Failed reading certificate file: " + getCertificatePath());
+        }
+    }
+
+    protected String readPrivateKeyFile() {
+        try {
+            return new String(Files.readAllBytes(Paths.get(getPrivateKeyPath())));
```
deepanjan90: The `openInput()` utility method helps resolve relative path
So in the configs rather than providing the full path relative to the configuration the path could be set
```suggestion
           return IoUtils.toString(openInput(getPrivateKeyPath()), UTF_8);
```

PR #139, opened 2020-03-26
Firewall to Computer Instance connection

### LINKED ISSUE #126: GCP: Firewall to Computer Instance connection
Currently instances do not have the ability to add tags which is the main form that it seems that firewall rules use to connect themselves to instances. 

Computer instances should be extended to include the ability to add tags.

Additionally, firewall rules [docs](https://gyro.dev/providers/google/compute/compute-firewall-rule.html) say that target-tags and target-source-account parameters can only be set if the firewall rule is egress, which means there is no way to limit ingress rules to only a single instance. If this restriction actually exists, it should be lifted, otherwise the docs should be updated to remove that line. 

I am basing my logic on needing these items on these docs:
https://cloud.google.com/vpc/docs/using-firewalls

AustinTB:
Fixes #126


src/main/java/gyro/google/compute/InstanceTags.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright 2019, Perfect Sense, Inc.
```
deepanjan90:
```suggestion
 * Copyright 2020, Perfect Sense, Inc.
```

src/main/java/gyro/google/compute/InstanceTags.java
```diff
@@ -0,0 +1,89 @@
...
+    public Tags copyTo() {
+        Tags tags = new Tags();
+
+        tags.setItems(
+            !getItems().isEmpty()
+                ? getItems()
+                : null
+        );
+        tags.setFingerprint(getFingerprint());
+        return tags;
```
deepanjan90:
```suggestion
        
        return tags;
```

src/main/java/gyro/google/compute/InstanceTags.java
```diff
@@ -0,0 +1,89 @@
...
+        return items;
+    }
+
+    public void setItems(List<String> items) {
+        this.items = items;
+    }
+
+    /**
+     * A string which is essentially a hash of the tags' contents and changes after every request to modify or update tags.
+     * An up-to-date fingerprint must always be provided in order to update or change tags.
```
deepanjan90: It is an output field so user is not supplying this so is redundant
```suggestion
```

src/main/java/gyro/google/compute/InstanceResource.java
```diff
@@ -309,6 +311,24 @@ public void setPrivateIp(String privateIp) {
...
+    /**
+     * The set of tags for the instance.
+     *
+     * @subresource gyro.google.compute.InstanceTags
+     */
+    @Updatable
+    public InstanceTags getTags() {
+        if (tags == null) {
+            tags = newSubresource(InstanceTags.class);
+        }
```
deepanjan90: Do we really need this, if tags are not present ?

beetlebugorg:
I think instance tags should just be a list of strings. If I'm reading this code correctly they'll look like this:

```
    instance-tags
        items: ['tag1', 'tag2']
    end
```

Other than matching the internal API, is there any reason it can't look like this:

```
    tags: ['tag1', 'tag2']
```

PR #143, opened 2020-03-27
GCP: Service Accounts and Service Account Keys

### LINKED ISSUE #127: GCP: Service Accounts
>A service account is a special kind of account used by an application or a virtual machine (VM) instance, not a person. Applications use service accounts to make authorized API calls.

>For example, a Compute Engine VM may run as a service account, and that account can be given permissions to access the resources it needs. This way the service account is the identity of the service, and the service account's permissions control which resources the service can access.

>A service account is identified by its email address, which is unique to the account.

In addition to creating this resource, google::compute-instance should be extended to allow these to be set, like instance-template is.

This uses the IAM Project - Service Accounts API:
https://cloud.google.com/iam/docs/reference/rest/v1/projects.serviceAccounts

Screen Shot 2020-03-13 at 11 08 13 AM

harjain99:
Fixes #127

### DIFF
--- a/src/main/java/gyro/google/compute/ComputeServiceAccount.java
+++ b/src/main/java/gyro/google/compute/ComputeServiceAccount.java
@@ -21,6 +21,7 @@ import java.util.List;
 
 import com.google.api.services.compute.model.ServiceAccount;
 import gyro.core.resource.Diffable;
+import gyro.core.validation.Required;
 import gyro.google.Copyable;
 import gyro.google.iam.ServiceAccountResource;
 import gyro.google.util.Utils;
@@ -32,8 +33,9 @@ public class ComputeServiceAccount extends Diffable implements Copyable<ServiceA
     private List<String> scopes;
 
     /**
-     * The service account resource.
+     * The service account.
      */
+    @Required
     public ServiceAccountResource getServiceAccount() {
         return serviceAccount;
     }
--- a/src/main/java/gyro/google/compute/ComputeServiceAccount.java
+++ b/src/main/java/gyro/google/compute/ComputeServiceAccount.java
@@ -22,22 +22,24 @@ import java.util.List;
 import com.google.api.services.compute.model.ServiceAccount;
 import gyro.core.resource.Diffable;
 import gyro.google.Copyable;
+import gyro.google.iam.ServiceAccountResource;
+import gyro.google.util.Utils;
 
 public class ComputeServiceAccount extends Diffable implements Copyable<ServiceAccount> {
 
-    private String email;
+    private ServiceAccountResource serviceAccount;
 
     private List<String> scopes;
 
     /**
-     * Email address of the service account.
+     * The service account resource.
      */
-    public String getEmail() {
-        return email;
+    public ServiceAccountResource getServiceAccount() {
+        return serviceAccount;
     }
 
-    public void setEmail(String email) {
-        this.email = email;
+    public void setServiceAccount(ServiceAccountResource serviceAccount) {
+        this.serviceAccount = serviceAccount;
     }
 
     /**
@@ -56,20 +58,20 @@ public class ComputeServiceAccount extends Diffable implements Copyable<ServiceA
 
     @Override
     public void copyFrom(ServiceAccount model) {
-        setEmail(model.getEmail());
+        setServiceAccount(findById(ServiceAccountResource.class, model.getEmail()));
         setScopes(model.getScopes());
     }
 
     public ServiceAccount toServiceAccount() {
         ServiceAccount serviceAccount = new ServiceAccount();
-        serviceAccount.setEmail(getEmail());
+        serviceAccount.setEmail(Utils.getServiceAccountEmailFromId(getServiceAccount().getId()));
         serviceAccount.setScopes(getScopes());
         return serviceAccount;
     }
 
     @Override
     public String primaryKey() {
-        String email = getEmail();
+        String email = Utils.getServiceAccountEmailFromId(getServiceAccount().getId() == null ? getServiceAccount().getName() : getServiceAccount().getId());
 
         if (email != null) {
             return email;
--- a/src/main/java/gyro/google/iam/RoleCustomProjectRoleFinder.java
+++ b/src/main/java/gyro/google/iam/RoleCustomProjectRoleFinder.java
@@ -34,7 +34,7 @@ import gyro.google.GoogleFinder;
  *
  * .. code-block:: gyro
  *
- *    role-custom-project: $(external-query google::iam-role-custom-project { name: 'projects/[PROJECT_NAME]/roles/role.custom.example' })
+ *    role-custom-project: $(external-query google::iam-role-custom-project { name: 'roles/role.custom.example' })
  */
 @Type("iam-role-custom-project")
 public class RoleCustomProjectRoleFinder extends GoogleFinder<Iam, Role, RoleCustomProjectRoleResource> {
@@ -80,7 +80,7 @@ public class RoleCustomProjectRoleFinder extends GoogleFinder<Iam, Role, RoleCus
         if (filters.containsKey("name")) {
             roles.add(client.projects()
                 .roles()
-                .get(filters.get("name")).execute());
+                .get("projects/" + getProjectId() + "/" + filters.get("name")).execute());;
         }
 
         return roles;
--- a/src/main/java/gyro/google/iam/RoleCustomProjectRoleFinder.java
+++ b/src/main/java/gyro/google/iam/RoleCustomProjectRoleFinder.java
@@ -80,7 +80,7 @@ public class RoleCustomProjectRoleFinder extends GoogleFinder<Iam, Role, RoleCus
         if (filters.containsKey("name")) {
             roles.add(client.projects()
                 .roles()
-                .get("projects/" + getProjectId() + "/" + filters.get("name")).execute());;
+                .get("projects/" + getProjectId() + "/" + filters.get("name")).execute());
         }
 
         return roles;
--- a/src/main/java/gyro/google/iam/RoleCustomProjectRoleFinder.java
+++ b/src/main/java/gyro/google/iam/RoleCustomProjectRoleFinder.java
@@ -34,7 +34,7 @@ import gyro.google.GoogleFinder;
  *
  * .. code-block:: gyro
  *
- *    role-custom-project: $(external-query google::iam-role-custom-project { name: 'roles/role.custom.example' })
+ *    role-custom-project: $(external-query google::iam-role-custom-project { name: 'projects/[PROJECT_NAME]/roles/role.custom.example' })
  */
 @Type("iam-role-custom-project")
 public class RoleCustomProjectRoleFinder extends GoogleFinder<Iam, Role, RoleCustomProjectRoleResource> {
@@ -80,7 +80,7 @@ public class RoleCustomProjectRoleFinder extends GoogleFinder<Iam, Role, RoleCus
         if (filters.containsKey("name")) {
             roles.add(client.projects()
                 .roles()
-                .get("projects/" + getProjectId() + "/" + filters.get("name")).execute());
+                .get(filters.get("name")).execute());
         }
 
         return roles;
--- a/src/main/java/gyro/google/iam/ServiceAccountFinder.java
+++ b/src/main/java/gyro/google/iam/ServiceAccountFinder.java
@@ -66,7 +66,11 @@ public class ServiceAccountFinder extends GoogleFinder<Iam, ServiceAccount, Serv
 
     @Override
     protected List<ServiceAccount> findAllGoogle(Iam client) throws Exception {
-        return client.projects().serviceAccounts().list(String.format("projects/%s", getProjectId())).execute().getAccounts();
+        return client.projects()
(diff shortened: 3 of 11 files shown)


src/main/java/gyro/google/iam/ServiceAccountKeyFinder.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Query service account keys.
+ *
+ * Example
+ * -------
+ *
+ * .. code-block:: gyro
+ *
+ *    service-account-key: $(external-query google::service-account-key {service-account-id: "projects/aerobic-lock-236714/serviceAccounts/example-service-account@aerobic-lock-236714.iam.gserviceaccount.com"})
```
deepanjan90: Do not include project sensitive info
```suggestion
 *    service-account-key: $(external-query google::service-account-key {service-account-id: ""})
```

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,254 @@
...
+
+        List<ServiceAccount> serviceAccounts = client.projects()
+            .serviceAccounts()
+            .list("projects/" + getProjectId())
+            .execute()
+            .getAccounts()
+            .stream()
+            .filter(r -> r.getName().equals(getId()))
+            .collect(
+                Collectors.toList());
```
deepanjan90: If you only care about one, then `findFirst()` is the way to go.
```suggestion
        ServiceAccount serviceAccount = client.projects()
            .serviceAccounts()
            .list("projects/" + getProjectId())
            .execute()
            .getAccounts()
            .stream()
            .filter(r -> r.getName().equals(getId()))
            .findfirst()
            .orElse(null);
```

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,254 @@
...
+        request.setServiceAccount(serviceAccount);
+
+        ServiceAccount response = client.projects()
+            .serviceAccounts()
+            .create("projects/" + getProjectId(), request)
+            .execute();
+
+        setId(response.getName());
+
+        changeServiceAccountStatus(client);
```
deepanjan90: check the status enable/disable and then fire this

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,254 @@
...
+        CreateServiceAccountRequest request = new CreateServiceAccountRequest();
+        request.setAccountId(getName());
+        request.setServiceAccount(serviceAccount);
+
+        ServiceAccount response = client.projects()
+            .serviceAccounts()
+            .create("projects/" + getProjectId(), request)
+            .execute();
+
+        setId(response.getName());
```
deepanjan90: So that if it fails in the next api call, corresponding gyro up can acknowledge that the resource was created
```suggestion
        setId(response.getName());
        state.save();
```

src/main/java/gyro/google/iam/ServiceAccountKeyFinder.java
```diff
@@ -0,0 +1,52 @@
...
+    protected List<ServiceAccountKey> findAllGoogle(Iam client) throws Exception {
+        throw new UnsupportedOperationException("Finding `service-account-keys` without filters is not supported!!");
+    }
+
+    @Override
+    protected List<ServiceAccountKey> findGoogle(Iam client, Map<String, String> filters) throws Exception {
+        List<ServiceAccountKey> keys = new ArrayList<>();
+
+        if (filters.containsKey("service-account-id")) {
+            keys = client.projects().serviceAccounts().keys().list(filters.get("service-account-id")).execute().getKeys();
```
deepanjan90: Can this not be executed without filters ?

src/main/java/gyro/google/iam/ServiceAccountFinder.java
```diff
@@ -0,0 +1,33 @@
...
+public class ServiceAccountFinder extends GoogleFinder<Iam, ServiceAccount, ServiceAccountResource> {
+
+    @Override
+    protected List<ServiceAccount> findAllGoogle(Iam client) throws Exception {
+        return client.projects().serviceAccounts().list("projects/" + getProjectId()).execute().getAccounts();
+    }
+
+    @Override
+    protected List<ServiceAccount> findGoogle(Iam client, Map<String, String> filters) throws Exception {
+        throw new UnsupportedOperationException("Finding `service-accounts` with filters is not supported!!");
```
deepanjan90: This should be supported accepting name and display name
Use the stream api if the api itself does not support this natively

src/main/java/gyro/google/iam/ServiceAccountKeyResource.java
```diff
@@ -0,0 +1,135 @@
...
+    @Override
+    protected boolean doRefresh() throws Exception {
+        Iam client = createClient(Iam.class);
+
+        List<ServiceAccountKey> keys = client.projects()
+            .serviceAccounts()
+            .keys()
+            .list(getServiceAccount().getId())
+            .execute()
+            .getKeys().stream().filter(r -> r.getName().equals(getId())).collect(Collectors.toList());
```
deepanjan90: Like the other one use `findFirst()` to only get one Item

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,254 @@
...
+public class ServiceAccountResource extends GoogleResource implements Copyable<ServiceAccount> {
+
+    private String displayName;
+    private String description;
+    private String name;
+    private Boolean enableAccount;
+
+    // Read-only
+    private String id;
+    private String email;
```
deepanjan90: How is this an output field, where does it get the email from ?

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,254 @@
...
+        setName(Utils.getServiceAccountNameFromId(model.getName()));
+        setDescription(model.getDescription());
+        setDisplayName(model.getDisplayName());
+        setEnableAccount((model.getDisabled() == null || model.getDisabled().equals(Boolean.FALSE)) ? Boolean.TRUE : Boolean.FALSE);
+    }
+
+    @Override
+    public String primaryKey() {
+        return super.primaryKey();
+    }
```
deepanjan90: why even bother mentioning this ?

src/main/java/gyro/google/util/Utils.java
```diff
@@ -66,6 +66,16 @@ public static String convertToFilters(Map<String, String> filterMap) {
             .collect(Collectors.joining(" "));
     }
 
+    public static String getServiceAccountIdFromName(String name, String projectId) {
+        return "projects/" + projectId + "/serviceAccounts/" + name + "@" + projectId + ".iam.gserviceaccount.com";
```
deepanjan90: No existing api to generate this ?

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,254 @@
...
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    /**
+     * Enable or disable the service account.
```
deepanjan90:
```suggestion
     * Enable or disable the service account. Defaults to ``????``.
```

src/main/java/gyro/google/util/Utils.java
harjain99: I wasn't able to find any such api.

src/main/java/gyro/google/iam/ServiceAccountResource.java
harjain99: The email gets set in the copyFrom. It's a part of the id and it uniquely identifies an account.

src/main/java/gyro/google/iam/ServiceAccountKeyFinder.java
harjain99: A Service Account ID is required to list its corresponding keys.

src/main/java/gyro/google/iam/ServiceAccountFinder.java
```diff
@@ -0,0 +1,72 @@
+package gyro.google.iam;
```
deepanjan90, line 1: Missing license text

src/main/java/gyro/google/iam/ServiceAccountKeyFinder.java
```diff
@@ -0,0 +1,52 @@
+package gyro.google.iam;
```
deepanjan90, line 1: Missing license text

src/main/java/gyro/google/util/Utils.java
```diff
@@ -66,6 +66,22 @@ public static String convertToFilters(Map<String, String> filterMap) {
...
```
deepanjan90, line 70: Use string format for better readability

RichieHowell:
#127 is not solved by this. The point of service accounts as supplied in that issue is to attach permissions to an instance so it can authenticate to GCP objects. I can't see a way in this PR to attach permissions to service accounts. In the console, this is done on the second tab on the creation:

Screen Shot 2020-04-02 at 9 20 07 AM


This solves half of my issue (that I cannot attach service accounts to instances), but does not solve the half where I have to manually attach roles to the service account. 


Separate note:

Instance Template currently allows you to attach service accounts but does it through supplying email of the service account. That should probably be updated in this change to be a service account resource like Instance was updated.

RichieHowell:
https://cloud.google.com/iam/docs/granting-roles-to-service-accounts  

It seems this is done by attaching IAM Policies to a service account.

IAM Policies and Roles should be added as resources if that is necessary to make this possible.

src/main/java/gyro/google/compute/ComputeServiceAccount.java
```diff
@@ -22,22 +22,24 @@
...
 public class ComputeServiceAccount extends Diffable implements Copyable<ServiceAccount> {
 
-    private String email;
+    private ServiceAccountResource serviceAccount;
 
     private List<String> scopes;
 
     /**
-     * Email address of the service account.
+     * The service account resource.
```
deepanjan90, line 35:
```suggestion
     * The service account.
```

src/main/java/gyro/google/compute/ComputeServiceAccount.java
```diff
@@ -22,22 +22,24 @@
...
      */
-    public String getEmail() {
-        return email;
+    public ServiceAccountResource getServiceAccount() {
```
deepanjan90, line 37:
```suggestion
    @Required
    public ServiceAccountResource getServiceAccount() {
```

src/main/java/gyro/google/iam/RoleCustomProjectRoleFinder.java
```diff
@@ -80,7 +80,7 @@ public void setName(String name) {
         if (filters.containsKey("name")) {
             roles.add(client.projects()
                 .roles()
-                .get("projects/" + getProjectId() + "/" + filters.get("name")).execute());
+                .get(filters.get("name")).execute());
```
deepanjan90, line 83: Why change this to explicitly provide a projectId ?

src/main/java/gyro/google/iam/ServiceAccountFinder.java
```diff
@@ -0,0 +1,88 @@
...
+        this.name = name;
+    }
+
+    public String getDisplayName() {
+        return displayName;
+    }
+
+    public void setDisplayName(String displayName) {
+        this.displayName = displayName;
+    }
```
deepanjan90, line 59: Missing java docs

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,386 @@
...
+            .execute();
+
+        setId(response.getName());
+        state.save();
+
+        if (getEnableAccount() != null && getEnableAccount().equals(Boolean.FALSE)) {
+            changeServiceAccountStatus(client);
+        }
+
+        copyFrom(response, false);
```
deepanjan90, line 234: Shouldn't copyFrom be called before state save to store output values

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,386 @@
...
+
+        if (!getPredefinedRoles().isEmpty() || !getCustomRoles().isEmpty()) {
```
deepanjan90, line 236: Another state save between these

src/main/java/gyro/google/iam/ServiceAccountResource.java
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    @Override
+    protected void doDelete(GyroUI ui, State state) throws Exception {
+        Iam client = createClient(Iam.class);
+
+        getCustomRoles().clear();
+        getPredefinedRoles().clear();
+
+        manageIamPolicies();
```
deepanjan90, line 274: Maybe a comment showing you are doing this to clear the iam roles
That is what this is doing right ?

PR #161, opened 2020-04-15
Set service account scope when linking with an instance

### LINKED ISSUE #160: Compute Instances : No way to set Service Account scope
The scope for service accounts is important when attaching a service account to an instance. Scopes further restrict access that the instance has when using the service account, and having the scope as null basically means that the instance has no access when using the service account, which is the same behavior as having no service account. 

https://cloud.google.com/compute/docs/access/service-accounts#accesscopesiam

There should be the ability to specify scopes when attaching service accounts to instances like there is when attaching to instance templates.

deepanjan90:
Fixes #160 

Service account resource was attached to instances as a list of direct reference.
But the attachment also allowed specifying the scope with every attachment.

Replacing the list of service account resource reference with the list of compute service account complex type [has a service account reference and a scope] solves this

### DIFF
--- a/src/main/java/gyro/google/compute/InstanceResource.java
+++ b/src/main/java/gyro/google/compute/InstanceResource.java
@@ -46,7 +46,6 @@ import gyro.core.validation.Required;
 import gyro.core.validation.ValidStrings;
 import gyro.core.validation.ValidationError;
 import gyro.google.Copyable;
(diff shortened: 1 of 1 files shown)


RichieHowell:
The list of service accounts on instance templates is serviceAccount and on instances it is serviceAccounts, they should match one way or another.

PR #162, opened 2020-04-16
Add option to auto generate disk name via initialize parameters

### LINKED ISSUE #154: Instance Templates and Instances : Make disk name not required.
On the docs here https://cloud.google.com/compute/docs/reference/rest/v1/instances and https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates, the disk name for initialize parameters say that if they are not specified, the instance name will be used instead. I need to use this for instance templates specifically as if I try to launch 4 autoscale instances, the 4th fails to start as it must share the same zone with another and fails to start due to boot disk name being the same as the other in the zone. 

The required tag should be removed from disk names on InstanceAttachedDiskInitializeParams.

deepanjan90:
Fixes #154


RichieHowell:
This worked for me, name was autogenerated on the instances created from the template.

PR #164, opened 2020-04-16
Backend Instances Health Status

### LINKED ISSUE #153: Compute Backends : Add health map method
Both the azure and aws providers include a way to get the number of healthy and unhealthy instances inside of Application Gateways and ELBs respectively. Compute Backends should be updated to include the same.

Links to the AWS/Azure provider equivalents:

https://github.com/perfectsense/gyro-azure-provider/blob/master/src/main/java/gyro/azure/network/ApplicationGatewayResource.java#L532-L564

https://github.com/perfectsense/gyro-aws-provider/blob/master/src/main/java/gyro/aws/elb/LoadBalancerResource.java#L258-L271

alex-lambrides:
Fixes #153

### DIFF
--- a/src/main/java/gyro/google/compute/BackendServiceResource.java
+++ b/src/main/java/gyro/google/compute/BackendServiceResource.java
@@ -16,17 +16,25 @@
 
 package gyro.google.compute;
 
+import java.io.IOException;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


deepanjan90:
Looks good 👍

PR #165, opened 2020-04-17
Add conditions while attaching roles

### LINKED ISSUE #163: Service Accounts : Allow setting conditions when attaching roles.
When attaching roles to service accounts in the console, you can define conditions which limit the service account's access to the role. We will be using this so that certain instances can only download storage objects they should have access to, and not have access to objects they shouldn't. 

Screen Shot 2020-04-16 at 1 10 47 PM

I believe that the binding used to attach roles to service accounts already has this method of setting, and we just need to allow it to be set through gyro.
https://developers.google.com/resources/api-libraries/documentation/cloudresourcemanager/v1/java/latest/com/google/api/services/cloudresourcemanager/model/Binding.html#setCondition-com.google.api.services.cloudresourcemanager.model.Expr-

harjain99:
Fixes #163

### DIFF
--- a/examples/iam/service-account.gyro
+++ b/examples/iam/service-account.gyro
@@ -10,11 +10,14 @@ google::service-account example-service-account
     name: "example-service-account-1"
     display-name: "example-service-account"
     description: "example service account"
+
     service-account-role
         predefined-role: $(external-query google::iam-role-predefined { name: "roles/viewer" })
     end
+
     service-account-role
         custom-role: $(google::iam-role-custom-project 'iam-role-custom-project-example')
+
         condition
             title: "example-condition"
             expression: 'resource.service != "iap.googleapis.com"'
--- a/examples/iam/service-account.gyro
+++ b/examples/iam/service-account.gyro
@@ -10,10 +10,16 @@ google::service-account example-service-account
     name: "example-service-account-1"
     display-name: "example-service-account"
     description: "example service account"
-    predefined-roles: $(external-query google::iam-role-predefined { name: "roles/viewer" })
-    custom-roles: [
-        $(google::iam-role-custom-project 'iam-role-custom-project-example')
-    ]
+    service-account-role
+        predefined-role: $(external-query google::iam-role-predefined { name: "roles/viewer" })
+    end
+    service-account-role
+        custom-role: $(google::iam-role-custom-project 'iam-role-custom-project-example')
+        condition
+            title: "example-condition"
+            expression: 'resource.service != "iap.googleapis.com"'
+        end
+    end
 end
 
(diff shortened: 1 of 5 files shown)


RichieHowell:
When trying to update a service account I created in the console before the ability to add in gyro existed, I got this 

```
⟳ Updating google::service-account gateway (projects/ops-test-270215/serviceAccounts/ops-test-gcp-network-1-gateway@ops-test-270215.iam.gserviceaccount.com) (change custom-roles)
Error: Can't update google::service-account gateway resource!

Caused by: Error: Specified policy version (1) cannot be less than the existing policy version (3). For more information, please refer to https://cloud.google.com/iam/docs/policies#versions.
```

The policy version should probably be grabbed from the current version and incremented.

RichieHowell:
New roles and service accounts work perfectly.

The output for replacing is a tad confusing:

Screen Shot 2020-04-22 at 11 11 54 AM

The bottom says it won't be replacing the expression but it does. If possible, that message should just be removed since it could indicate to end user that it won't change when it will. To be clear, the update works as intended and the next gyro up gives "No Changes" as expected, the messaging around the original change is just a tad confusing.

harjain99:
> When trying to update a service account I created in the console before the ability to add in gyro existed, I got this
> 
> ```
> ⟳ Updating google::service-account gateway (projects/ops-test-270215/serviceAccounts/ops-test-gcp-network-1-gateway@ops-test-270215.iam.gserviceaccount.com) (change custom-roles)
> Error: Can't update google::service-account gateway resource!
> 
> Caused by: Error: Specified policy version (1) cannot be less than the existing policy version (3). For more information, please refer to https://cloud.google.com/iam/docs/policies#versions.
> ```
> 
> The policy version should probably be grabbed from the current version and incremented.

https://github.com/perfectsense/gyro-google-provider/pull/165/files#diff-6cfb9d3cc8a9caf736e7c403b57b939aR387 makes sure that every time a policy is retrieved, the requested version is set to 3. Lower versions do not support managing conditions.

harjain99:
> New roles and service accounts work perfectly.
> 
> The output for replacing is a tad confusing:
> 
> Screen Shot 2020-04-22 at 11 11 54 AM
> 
> The bottom says it won't be replacing the expression but it does. If possible, that message should just be removed since it could indicate to end user that it won't change when it will. To be clear, the update works as intended and the next gyro up gives "No Changes" as expected, the messaging around the original change is just a tad confusing.

Fixed in https://github.com/perfectsense/gyro-google-provider/pull/165/commits/a54f4e1edcfcecf8be68aec2be1238bc6a9a4afc

RichieHowell:
output is much nicer now, this fixes all my concerns.

PR #191, opened 2020-07-09
Enables Remote State Storage

### LINKED ISSUE #190: Allow GCP Storage to work with remote state backends
We need to implement `FileBackend#list` in order to allow state to be stored on GCP Storage.

alex-lambrides:
Fixes #190

### DIFF
--- a/src/main/java/gyro/google/GoogleStorageFileBackend.java
+++ b/src/main/java/gyro/google/GoogleStorageFileBackend.java
@@ -19,15 +19,17 @@ package gyro.google;
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


src/main/java/gyro/google/GoogleStorageFileBackend.java
```diff
@@ -69,7 +71,34 @@ public String getCredentials() {
...
+            Storage client = client();
+            String nextPageToken = null;
+            int count = 0;
+
+            do {
+                Objects objects = client.objects()
+                    .list(getBucket())
+                    .setPrefix(prefixed(""))
+                    .setPageToken(nextPageToken)
+                    .execute();
```
deepanjan90, line 85: This should be paginated just like we have a limit of 10k for the aws variant

alex-lambrides: Addressed with Andrew's pr: #192

PR #240, opened 2022-03-21
Fix setting router nat on create

deepanjan90:
Fixes the issue of re gyro up when setting nat router for a nat.


deepanjan90:
Already gets addressed https://github.com/perfectsense/gyro-google-provider/pull/237

PR #244, opened 2022-07-26
Service account refresh issue

deepanjan90:
Gyro sometimes is unable to find a service account on refresh.

Cause gyro to want to create. a service account it just created.

This occurs due to the api returning paged result of listing service accounts  and gyro only looking at the first page.


harjain99 (approved):
reproduced, and tested the fix.

Issue #249, opened 2023-05-22
Support Cloud Armor


tloisel1:
https://cloud.google.com/armor

Cloud Armor is GCP's WAF service. We should be able to control this through Gyro

PR #250, opened 2023-05-22
Add additional options for cloud armor

deepanjan90:
Fixes: #249 https://perfectsense.atlassian.net/browse/INF-252

Added support for
 - advanced option [ Not supported on UI yet ]
 - adaptive protection
 
 - Updated rule option
   - redirect
   - throttle
   - rate-based-ban
   - Added support for headers in allow/deny

### DIFF
--- a/src/main/java/gyro/google/compute/SecurityPolicyRule.java
+++ b/src/main/java/gyro/google/compute/SecurityPolicyRule.java
@@ -16,6 +16,8 @@
 
 package gyro.google.compute;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 
 import com.google.cloud.compute.v1.AddRuleSecurityPolicyRequest;
@@ -29,7 +31,9 @@ import gyro.core.resource.Updatable;
 import gyro.core.scope.State;
 import gyro.core.validation.Required;
 import gyro.core.validation.ValidStrings;
+import gyro.core.validation.ValidationError;
 import gyro.google.Copyable;
+import org.apache.commons.lang3.StringUtils;
 
 public class SecurityPolicyRule extends ComputeResource
     implements Copyable<com.google.cloud.compute.v1.SecurityPolicyRule> {
@@ -39,6 +43,9 @@ public class SecurityPolicyRule extends ComputeResource
     private String action;
     private Boolean preview;
     private SecurityPolicyRuleMatcher match;
+    private SecurityPolicyRuleHttpHeaderActionConfig headerAction;
+    private SecurityPolicyRuleRedirectOptionsConfig redirectConfig;
+    private SecurityPolicyRuleRateLimitOptionsConfig rateLimitConfig;
 
     /**
      * The description of the security policy rule.
@@ -69,7 +76,7 @@ public class SecurityPolicyRule extends ComputeResource
      */
     @Updatable
     @Required
-    @ValidStrings({ "allow", "deny(403)", "deny(404)", "deny(502)" })
+    @ValidStrings({ "allow", "deny(403)", "deny(404)", "deny(502)", "rate_based_ban", "redirect", "throttle" })
     public String getAction() {
         return action;
     }
@@ -79,10 +86,14 @@ public class SecurityPolicyRule extends ComputeResource
     }
 
     /**
-     * The preview flag indicates that this rule is not enforced.
+     * The preview flag indicates that this rule is not enforced. Defaults to ``false``.
      */
     @Updatable
     public Boolean getPreview() {
+        if (preview == null) {
+            preview = false;
+        }
+
         return preview;
     }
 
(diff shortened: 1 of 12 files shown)


src/main/java/gyro/google/compute/SecurityPolicyRuleRateLimitOptionsConfig.java
```diff
@@ -0,0 +1,302 @@
...
+
+    // TODO: Not supported by the api yet
+    private String enforceOnKey;
+    private String enforceOnKeyName;
+
+    // TODO: Not supported in the UI yet
+    private SecurityPolicyRuleRedirectOptionsConfig exceedRedirectConfig;
+
+    /**
+     * The ban threshold config. Only valid when rule `action` is set to `rate_based_ban`.
```
harjain99, line 50:
```suggestion
     * The ban threshold config. Only valid when rule ``action`` is set to ``rate_based_ban``.
```

src/main/java/gyro/google/compute/SecurityPolicyRuleRedirectOptionsConfig.java
```diff
@@ -0,0 +1,102 @@
...
+        return type;
+    }
+
+    public void setType(String type) {
+        this.type = type;
+    }
+
+    @Override
+    public String primaryKey() {
+        return null;
```
harjain99, line 66:
```suggestion
        return "";
```

src/main/java/gyro/google/compute/SecurityPolicyRule.java
```diff
@@ -105,9 +116,52 @@ public void setMatch(SecurityPolicyRuleMatcher match) {
...
+
+    public void setRateLimitConfig(SecurityPolicyRuleRateLimitOptionsConfig rateLimitConfig) {
+        this.rateLimitConfig = rateLimitConfig;
+    }
+
     @Override
     public String primaryKey() {
-        return "with priority " + getPriority();
+        return String.format("%swith priority %s",
+            (StringUtils.isBlank(getDescription()) ? "" : String.format("(%s) ", getDescription())), getPriority());
```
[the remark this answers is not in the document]
deepanjan90: The `description` along with other fields can get updated without creating a new sub resource, but given that there can be multiple rules and no same priority can be assigned, we chose to keep the `priority` in the primary key

src/main/java/gyro/google/compute/SecurityPolicyRuleRateLimitOptionsConfig.java
```diff
@@ -0,0 +1,302 @@
...
+        this.exceedAction = exceedAction;
+    }
+
+    /**
+     * The key to enforce the rate limit on.
+     */
+    @Updatable
+    public String getEnforceOnKey() {
+        return enforceOnKey;
+    }
```
harjain99, line 127:
```suggestion
    /**
     * The key to enforce the rate limit on.
     */
    @Updatable
    @ValidStrings({ "ALL", "IP", "HTTP_HEADER", "XFF_IP", "HTTP_COOKIE", "HTTP_PATH", "SNI", "REGION_CODE" })
    public String getEnforceOnKey() {
        return enforceOnKey;
    }
```


Has the particular valid Strings:

Determines the key to enforce the rate_limit_threshold on. Possible values are:
- ALL: A single rate limit threshold is applied to all the requests matching this rule. This is the default value if "enforceOnKey" is not configured.
- IP: The source IP address of the request is the key. Each IP has this limit enforced separately.
- HTTP_HEADER: The value of the HTTP header whose name is configured under "enforceOnKeyName". The key value is truncated to the first 128 bytes of the header value. If no such header is present in the request, the key type defaults to ALL.
- XFF_IP: The first IP address (i.e. the originating client IP address) specified in the list of IPs under X-Forwarded-For HTTP header. If no such header is present or the value is not a valid IP, the key defaults to the source IP address of the request i.e. key type IP.
- HTTP_COOKIE: The value of the HTTP cookie whose name is configured under "enforceOnKeyName". The key value is truncated to the first 128 bytes of the cookie value. If no such cookie is present in the request, the key type defaults to ALL.
- HTTP_PATH: The URL path of the HTTP request. The key value is truncated to the first 128 bytes.
- SNI: Server name indication in the TLS session of the HTTPS request. The key value is truncated to the first 128 bytes. The key type defaults to ALL on a HTTP session.
- REGION_CODE: The country/region from which the request originates.
  Check the EnforceOnKey enum for the list of possible values.

src/main/java/gyro/google/compute/SecurityPolicyRuleRateLimitOptionsConfig.java
```diff
@@ -0,0 +1,302 @@
...
+        this.enforceOnKey = enforceOnKey;
+    }
+
+    /**
+     * The name of the key to enforce the rate limit on.
+     */
+    @Updatable
+    public String getEnforceOnKeyName() {
+        return enforceOnKeyName;
+    }
```
harjain99, line 139: Can only be set if `enforceOnKey` is set to `HTTP_COOKIE` or `HTTP_HEADER`.

harjain99 (changes requested):
The expression config doesn't seem to save the location and description and asks to update every time.

```
= Keep google::compute-security-policy security-policy-example (https://www.googleapis.com/compute/v1/projects/aerobic-lock-236714/global/securityPolicies/security-policy-example)
    ⟳ Update rule (redirect-rule-external-address-example) with priority 6
      · match:  match  → match 
        ⟳ Update match
          · expression-config:  expression-config  → expression-config 
            ⟳ Update expression-config
              · description:  '' → 'description'
              · location:  '' → '12'
```

harjain99 (approved):

PR #274, opened 2024-09-23
Implement cloudsql

harjain99:


src/main/java/gyro/google/cloudsql/DatabaseInstanceResource.java
```diff
@@ -0,0 +1,950 @@
...
+    }
+
+    @Override
+    protected void doCreate(GyroUI ui, State state) throws Exception {
+        SQLAdmin client = createClient(SQLAdmin.class);
+        DatabaseInstance databaseInstance = new DatabaseInstance();
+
+        if (getReplicationCluster() != null) {
+            throw new GyroException(
+                "The 'replication-cluster' field is not supported for creating a new instance. It can only be configured in an update call.");
```
deepanjan90, line 787: Rather than throwing the error, we should follow up with an update call to handle this, rather than making the user do gyro up again

src/main/java/gyro/google/cloudsql/DatabaseInstanceResource.java
```diff
@@ -0,0 +1,950 @@
...
+
+        if (getSettings() != null) {
+            databaseInstance.setSettings(getSettings().toSettings());
+        }
+
+        if (getSwitchTransactionLogsToCloudStorageEnabled() != null) {
+            databaseInstance.setSwitchTransactionLogsToCloudStorageEnabled(getSwitchTransactionLogsToCloudStorageEnabled());
+        }
+
+        client.instances().insert(getProjectId(), databaseInstance);
```
deepanjan90, line 868: are there other resources that might need this resource ?
Asking to know if a thought to a wait function was given

src/main/java/gyro/google/cloudsql/DbAclEntry.java
```diff
@@ -0,0 +1,95 @@
...
+        setValue(model.getValue());
+    }
+
+    @Override
+    public String primaryKey() {
+        return String.format(
+            "Acl Entry [Value: %s, Expiration Time: %s, Name: %s]",
+            getValue(),
+            getExpirationTime(),
+            getName());
```
deepanjan90, line 77: This will look funky if only the value is provided.
We should probably create this string conditionally

src/main/java/gyro/google/cloudsql/DbBackupConfiguration.java
```diff
@@ -0,0 +1,179 @@
...
+
+    public void setPointInTimeRecoveryEnabled(Boolean pointInTimeRecoveryEnabled) {
+        this.pointInTimeRecoveryEnabled = pointInTimeRecoveryEnabled;
+    }
+
+    /**
+     * The start time for the daily backup configuration in UTC timezone in the 24 hour format - `HH:MM`.
+     */
+    @Updatable
+    @Regex("[0-2][0-9]:[0-5][0-9]")
```
deepanjan90, line 106: the hour regex will fail, 25-29 will be valid by this regex

src/main/java/gyro/google/cloudsql/DbBackupConfiguration.java
```diff
@@ -0,0 +1,179 @@
...
+        setEnabled(model.getEnabled());
+        setLocation(model.getLocation());
+        setPointInTimeRecoveryEnabled(model.getPointInTimeRecoveryEnabled());
+        setStartTime(model.getStartTime());
+        setTransactionLogRetentionDays(model.getTransactionLogRetentionDays());
+        setTransactionalLogStorageState(model.getTransactionalLogStorageState());
+
+        setBackupRetentionSettings(null);
+        if (model.getBackupRetentionSettings() != null) {
+            DbBackupRetentionSettings settings = new DbBackupRetentionSettings();
```
deepanjan90, line 152: new subresource ?

src/main/java/gyro/google/cloudsql/DbInsightsConfig.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Override
+    public String primaryKey() {
+        return "";
+    }
+
+    public InsightsConfig toInsightsConfig() {
+        InsightsConfig config = new InsightsConfig();
+
+        if (getQueryInsightsEnabled() != null) {
```
deepanjan90, line 114: Given this field is a required value, this will never be a null conditional right ?

src/main/java/gyro/google/cloudsql/DbIpConfiguration.java
```diff
@@ -0,0 +1,212 @@
...
+    public String getAllocatedIpRange() {
+        return allocatedIpRange;
+    }
+
+    public void setAllocatedIpRange(String allocatedIpRange) {
+        this.allocatedIpRange = allocatedIpRange;
+    }
+
+    /**
+     * The list of external networks that are allowed to connect to the instance using the IP.
```
deepanjan90, line 54: doc entry with `@subresource`

src/main/java/gyro/google/cloudsql/DbLocationPreference.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public LocationPreference toLocationPreference() {
+        LocationPreference preference = new LocationPreference();
+
+        if (getSecondaryZone() != null) {
+            preference.setSecondaryZone(getSecondaryZone());
+        }
+        if (getZone() != null) {
+            preference.setZone(getZone());
+        }
```
deepanjan90, line 74: None of them are required, that will create a blank class ?

src/main/java/gyro/google/cloudsql/DbMySqlReplicaConfiguration.java
```diff
@@ -0,0 +1,182 @@
...
+ * limitations under the License.
+ */
+
+package gyro.google.cloudsql;
+
+import com.google.api.services.sqladmin.model.MySqlReplicaConfiguration;
+import gyro.core.resource.Diffable;
+import gyro.google.Copyable;
+
+public class DbMySqlReplicaConfiguration extends Diffable implements Copyable<MySqlReplicaConfiguration> {
```
deepanjan90, line 23: Any required/updatable fields ?

src/main/java/gyro/google/cloudsql/DbOnPremisesConfiguration.java
```diff
@@ -0,0 +1,192 @@
...
+ */
+
+package gyro.google.cloudsql;
+
+import com.google.api.services.sqladmin.model.InstanceReference;
+import com.google.api.services.sqladmin.model.OnPremisesConfiguration;
+import gyro.core.resource.Diffable;
+import gyro.google.Copyable;
+
+public class DbOnPremisesConfiguration extends Diffable implements Copyable<OnPremisesConfiguration> {
```
deepanjan90, line 24: Any required/updatable fields ?

src/main/java/gyro/google/cloudsql/DbPasswordValidationPolicy.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Override
+    public String primaryKey() {
+        return "";
+    }
+
+    public PasswordValidationPolicy toPasswordValidationPolicy() {
+        PasswordValidationPolicy policy = new PasswordValidationPolicy();
+        policy.setEnablePasswordPolicy(getEnablePasswordPolicy());
```
deepanjan90, line 127: Is the rest of the configuration still allowed if this is set to false ?

src/main/java/gyro/google/cloudsql/DbPscConfig.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    @Override
+    public String primaryKey() {
+        return "";
+    }
+
+    public PscConfig toPscConfig() {
+        PscConfig config = new PscConfig();
+        config.setPscEnabled(getPscEnabled());
```
deepanjan90, line 77: if this is set to false can you still set the projects ?

src/main/java/gyro/google/cloudsql/DbReplicationCluster.java
```diff
@@ -0,0 +1,91 @@
...
+    private String psaWriteEndpoint;
+
+    /**
+     * The disaster recovery (DR) replica for a primary instance.
+     * If the instance is a read replica, then the field is not set. Set this field to a replica name to designate a DR
+     * replica for a primary instance. Remove the replica name to remove the DR replica designation.
+     * Only for Enterprise Plus edition instances.
+     */
+    @Updatable
+    public String getFailoverDrReplicaName() {
```
deepanjan90, line 40: This will probably  be required, given it is the only settable field in this class

src/main/java/gyro/google/cloudsql/DbSettings.java
```diff
@@ -0,0 +1,727 @@
...
+    }
+
+    @Override
+    public String primaryKey() {
+        return "";
+    }
+
+    public Settings toSettings() {
+        Settings settings = new Settings();
+        settings.setSettingsVersion(getSettingsVersion());
```
deepanjan90, line 595: This will probably be a required field then.

Also given this is a db settings, re check to see if there are other fields that are required.

src/main/java/gyro/google/cloudsql/DbSqlScheduledMaintenance.java
```diff
@@ -0,0 +1,87 @@
...
+    public String primaryKey() {
+        return "";
+    }
+
+    public SqlScheduledMaintenance toSqlScheduledMaintenance() {
+        SqlScheduledMaintenance config = new SqlScheduledMaintenance();
+
+        config.setCanReschedule(getCanReschedule());
+        config.setScheduleDeadlineTime(getScheduleDeadlineTime());
+        config.setStartTime(getStartTime());
```
deepanjan90, line 83: Can all of these be set to null, given none of this is required

deepanjan90 (changes requested):
All of the copyFrom() should have suberesource object created with `newSubresource` declarations.

Double check all the docs for subresource to make sure they have the doc entry with `@subresource`

src/main/java/gyro/google/cloudsql/DbLocationPreference.java
harjain99: The api lets you set these to null and still has default values for them.

src/main/java/gyro/google/cloudsql/DbMySqlReplicaConfiguration.java
harjain99: Doesn't seem like it from the sdk

src/main/java/gyro/google/cloudsql/DbOnPremisesConfiguration.java
harjain99: Doesn't seem like it from the sdk

src/main/java/gyro/google/cloudsql/DbPasswordValidationPolicy.java
harjain99: Added validation

src/main/java/gyro/google/cloudsql/DbPscConfig.java
harjain99: Added Validation

src/main/java/gyro/google/cloudsql/DbSettings.java
harjain99: The settings version is an output field so I don't think it's need the annotation.
It just needs to be provided on every update call.

The tier is basically the only thing required other than the name.

src/main/java/gyro/google/cloudsql/DbSqlScheduledMaintenance.java
harjain99: The api doesn't seem to complain even if they are all set to null values.

src/main/java/gyro/google/cloudsql/DatabaseInstanceResource.java
harjain99: So that problem here is that there is a sort of circular dependency. To set this value, a replica of the primary instance (which is basically another `DatabaseInstanceResource`) needs to be created first which then needs to be added back to the primary instance.
I don't think an update call would solve the issue here unfortunately.

src/main/java/gyro/google/cloudsql/DatabaseInstanceResource.java
harjain99: Added waiters

src/main/java/gyro/google/cloudsql/DbAclEntry.java
harjain99: Added a stringbuilder to construct the key

src/main/java/gyro/google/cloudsql/DbBackupConfiguration.java
harjain99: Fixed

src/main/java/gyro/google/cloudsql/DbInsightsConfig.java
harjain99: Fixed

deepanjan90 (approved):

PR #281, opened 2026-07-08
Support spot vms types for gke nodes

harjain99:
Adds ability to support spot vm types for gke nodes


ianp94 (approved):
Adds support for spot instances over preemptible. Looks good to me

